PR #3, opened 2016-12-02
Master

l-mas:
work done so far, might need more documentation


apdibbo:
What does this do?

l-mas:
it monitors the rabbitmq queues and attempts to run the aquilon commands when a vm is created/deleted. was this not the right repository to push to?

PR #15, opened 2017-06-30
Frazer update

frazzleuk:
I have updated the Aquilon Hook code such that it should be more stable. If it fails, it will fail in a detectable way and more verbally such that it is easier to track down what went wrong.


apdibbo:
LGTM @jrha would you mind weighing in on how the Aquilon API interaction looks?

Tom-Byrne:
I haven't had an exhaustive look through the code changes, but it looks good to me, seems to be working well, and the more verbose error messages have already been useful.

PR #18, opened 2017-11-02
OpenStack_get_username: Address issues flagged by shell check

jrha:
Use bash-isms, quote to prevent globbing and splitting, etc.
Remove commented out lines


apdibbo:
LGTM

PR #21, opened 2020-08-13
Add CGI script to map from project ID to IAM group

willfurnell:


willfurnell:
@apdibbo Is this all OK with you? Any changes needed or can it be put as the live version please?

Thank you!

PR #24, opened 2022-10-10
Add scripts for generating accounting information

apdibbo:
Add script for generating accounting information

### DIFF
--- a/OpenStack-nova-accounting/extract_accounting.py
+++ b/OpenStack-nova-accounting/extract_accounting.py
@@ -1,0 +1,115 @@
+#!/usr/bin/python
+import time
+import datetime
+#from datetime import datetime,time
+import json
+import requests
+import sys
+import sqlalchemy
+from sqlalchemy.sql import select
+from sqlalchemy.orm import sessionmaker
+from ConfigParser import SafeConfigParser
+
+
+# Read from config file
+parser = SafeConfigParser()
+try:
+   parser.read('/etc/influxdb.conf')
+   host = parser.get('db', 'host')
+   database = parser.get('db', 'database')
+   username = parser.get('auth', 'username')
+   password = parser.get('auth', 'password')
+   instance = parser.get('cloud','instance')
+except:
+   print 'Unable to read from config file'
+   sys.exit(1)
+
+url = 'http://'+host+'/write?db='+database +'&precision=s'
+
+nowtime = time.localtime()
+
+try:
+   parser.read('/etc/nova/nova.conf')
+   connectionstring = parser.get('database','connection')
+except:
+   print 'Unable to read from config file'
+   sys.exit(1)
+
+starttime='2017-04-19 12:00'
+starttime=sys.argv[1]
+print(starttime)
+endtime='2017-04-19 12:15'
+endtime=sys.argv[2]
+print(endtime)
+endyyyymm=datetime.datetime.strptime(endtime,"%Y-%m-%d %H:%M").strftime('%Y-%m')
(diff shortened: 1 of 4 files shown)


OpenStack-nova-accounting/extract_accounting.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/python
```
DavidFair, line 1: Could we explicitly mark this as Python 3 if it's compatible, to save me checking it later?

OpenStack-nova-accounting/extract_accounting.py
```diff
@@ -0,0 +1,115 @@
...
+
+url = 'http://'+host+'/write?db='+database +'&precision=s'
+
+nowtime = time.localtime()
+
+try:
+   parser.read('/etc/nova/nova.conf')
+   connectionstring = parser.get('database','connection')
+except:
+   print 'Unable to read from config file'
```
DavidFair, line 35: Use
`raise RuntimeError("Unable to read from config file")`
That way it will log to syserr and exit at the same time

OpenStack-nova-accounting/extract_accounting.py
```diff
@@ -0,0 +1,115 @@
...
+   sys.exit(1)
+
+starttime='2017-04-19 12:00'
```
DavidFair, line 38: This can be removed or turned into a comment if it's for reference

OpenStack-nova-accounting/extract_accounting.py
```diff
@@ -0,0 +1,115 @@
...
+starttime=sys.argv[1]
+print(starttime)
+endtime='2017-04-19 12:15'
```
DavidFair, line 41: ^

OpenStack-nova-accounting/extract_accounting.py
```diff
@@ -0,0 +1,115 @@
...
+
+    datastring += " "+str(long(endtimestamp))
+    datastring += "\n"
+
+print datastring
+
+
+r = requests.post(url,data=datastring,auth=(username,password))
+print r.text
+print r
```
DavidFair, line 115: I think these prints need to be wrapped and it's Python 3 ready

DavidFair (changes requested):

OpenStack-nova-accounting/extract_accounting.py
apdibbo: Done

OpenStack-nova-accounting/extract_accounting.py
apdibbo: Done

OpenStack-nova-accounting/extract_accounting.py
apdibbo: Done

OpenStack-nova-accounting/extract_accounting.py
apdibbo: Done

OpenStack-nova-accounting/extract_accounting.py
apdibbo: I've added comments about what needs to be done

OpenStack-nova-accounting/extract_accounting.py
```diff
@@ -21,7 +21,7 @@
    password = parser.get('auth', 'password')
    instance = parser.get('cloud','instance')
 except:
-   print('Unable to read from config file')
+   raise RuntimeError('Unable to read from config file')
    sys.exit(1)
```
DavidFair, line 27: Nit: these `sys.exit(1)` can be dropped as it's effectively dead code
But it's not worth holding up the PR for

DavidFair (approved):

PR #30, opened 2022-11-09
Add tests for methods in message_consumer

DavidFair:
Adds tests for methods in message_consumer 
Refactoring will come in the next PR, as we're already ~1k LOC churn


jacob-ward (approved):
Ideally I'd like more comments, but as most of it is self explanatory it's alright.

PR #31, opened 2022-11-22
Dockerise consumer

DavidFair:
- Creates the various docker files required for the app including moving the entrypoint to avoid a circular import
- Adds CI for the new docker file
- Cleans the legacy unit files and docs

To test:
- `cd OpenStack-Rabbit-Consumer`
- `cp prod.env .env`  # Note credentials don't need to be provided at this phase
- `docker build . -t stfc/rabbit-consumer:latest && docker-compose up`

The service should start, then fail to connect to the broker without credentials


apdibbo:
Looks fine to me but some of the tests did not complete?

PR #36, opened 2023-01-31
Rabbit consumer: Add steps to automatically update image on Harbor

DavidFair:
Adds steps to stage the image automatically for staging / qa


.github/workflows/rabbit_consumer.yaml
```diff
@@ -1,13 +1,13 @@
...
+on: [push, pull_request]
 
 jobs:
-  build:
+  test_and_lint:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.10"]
```
meoflynn, line 10: Why have we removed python 3.8 and 3.9 from the list of python versions here?

DavidFair: We only build the docker file against 3.10, so it's wasted checking these older versions which we won't run against

Good question :+1:

meoflynn (approved):

meoflynn:
@DavidFair there's a merge conflict, could you please have a look?

meoflynn (approved):

PR #37, opened 2023-01-31
Hardcode nova exchange

DavidFair:
We only read one exchange, since the env's should always been the same apply YAGNI and massively simplify the code. In the future if we add other exchange support we simply add to the hard-coded list since then envs should line up (or we add support to switch in-code)


OpenStack-Rabbit-Consumer/consumer.ini
```diff
@@ -1,2 +1,2 @@
 [rabbit]
-exchanges=nova
+exchanges=
```
meoflynn, line 2: Why is `exchanges=` ?

meoflynn (changes requested):

OpenStack-Rabbit-Consumer/consumer.ini
DavidFair: This should have been removed instead of just the text :man_facepalming:

meoflynn (approved):

PR #41, opened 2023-02-16
Add Git Operations for future CAPI Image Builder

DavidFair:
Adds the framework for handling the complex git ops we're going to do later on and associated unit tests

### DIFF
--- a/capi_image_builder/main.py
+++ b/capi_image_builder/main.py
@@ -1,0 +1,6 @@
+# Clone repo
+# Rebase
+# Remove old files
+# Build
+# Push to OpenStack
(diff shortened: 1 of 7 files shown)


capi_image_builder/main.py
```diff
@@ -0,0 +1,6 @@
+# Clone repo
```
meoflynn, line 1: Is this a placeholder for developing the `main.py` in a future PR?

DavidFair: Yes, it's the steps (IMO) we need to write out and add unit tests for

meoflynn (approved):

PR #46, opened 2023-03-09
Script that generates Prometheus config including IP list

RJaliliSTFC:
Changed to python and git for more universal practices, simplified the overall configuration


meoflynn (changes requested):
Looks good, please could we add descriptions to the methods and to the unit tests?

prometheus_ip_script/test/test_generate_ips.py
```diff
@@ -0,0 +1,82 @@
...
+    substitute_template,
+    write_to_file,
+)
+
+
+def test_generate_100():
+    """
+    Test that the generate_hosts function returns a list of 255 hosts
+    belonging to the 172.16.100.x network
+    :return:
```
meoflynn, line 17: Could we remove this line if this unit test does not return any values?

meoflynn (approved):

PR #49, opened 2023-03-21
Cloud energy collection 

anish-mudaraddi:
Draft adding iriscasttools - a python package to get energy usage metrics from cloud hypervisors and storage nodes.
Collects a variety of information including:
- instantaneous power in Watts
- Os load 
- RAM usage

Python package called by a cron job every 5 minutes on each hypervisor to collect data over time


anish-mudaraddi:
This tool makes use of `subprocess.popen` to call `ipmitool` to get power information.

There exists other python packages that aim to provide ipmi functionality as a python package

- [python-ipmi](https://github.com/kontron/python-ipmi)
- [hyve-pyipmi](https://github.com/hyvedesignsolutions/hyve-pyipmi)
 
Neither seem to allow us to execute dcmi commands - needed to get power information

Both seem to require a username and password for accessing ipmitool which we don't require here since the command is run locally

So I think the current approach is fine as is

.github/workflows/iriscast_package_build.yaml
```diff
@@ -0,0 +1,36 @@
...
+    paths:
+      - "cloud-energy-collection/**"
+      - ".github/workflows/iriscast_package_build.yaml"
+
+jobs:
+  test_and_lint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10"]
```
DavidFair, line 17: We should also test against the version of Python shipped in RL8 too

cloud-energy-collection/iriscasttools/main.py
```diff
@@ -0,0 +1,27 @@
+"""
+Collects energy usage metrics using IPMI, as well as other metrics such as CPU and RAM usage.
+"""
+import utils
+
+
+def get_iriscast_stats(csv=False, include_header=False):
```
DavidFair, line 7: This should be wrapped in a `if __main__` too:
https://stackoverflow.com/questions/419163/what-does-if-name-main-do

Like:
```
if __main__:
   get_iriscast_stats(...)
```

cloud-energy-collection/iriscasttools/test/test_main.py
```diff
@@ -0,0 +1,56 @@
...
+
+    expected_all = dict(expected_power_stats)
+    expected_all.update(expected_os_load_stats)
+    expected_all.update(expected_ram_stats)
+
+    mock_get_ipmi_power_stats.side_effect = [expected_power_stats]
+    mock_get_os_load.return_value = expected_os_load_stats
+    mock_get_ram_usage.return_value = expected_ram_stats
+
+    _ = get_iriscast_stats(test_csv_flag, test_include_header)
```
DavidFair, line 47: We should check that what `get_iriscast_stats` returns was the same as what one of the mock methods returned

Currently you could change it to `return None` and the test would still pass

cloud-energy-collection/iriscasttools/test/test_utils.py
```diff
@@ -0,0 +1,265 @@
...
+
+@pytest.mark.parametrize(
+    "valid_path",
+    [
+        ("/dev/ipmi0"),
+        ("/dev/ipmi/0"),
+        ("/dev/ipmidev/0"),
+    ],
+)
+def test_check_ipmi_conn(valid_path):
```
DavidFair, line 125: Do we have a test for if these paths don't exist? I.e. if someone accidentally runs this on a VM

cloud-energy-collection/iriscasttools/test/test_utils.py
```diff
@@ -0,0 +1,265 @@
...
+    expected_results.update(
+        {
+            k: v
+            for k, v in all_expected_vals_dict.items()
+            if k in expected_results.keys()
+        }
+    )
+
+    result = get_ipmi_power_stats(*test_args)
+    assert mock_check_ipmi_conn.call_count == 1
```
DavidFair, line 201: Nit: This can simply be `mock_check_ipmi_conn.assert_called_once()`
As it's more Pythonic

cloud-energy-collection/iriscasttools/utils.py
```diff
@@ -0,0 +1,208 @@
...
+        for line in res.splitlines():
+            line_str = line.split(":")
+            stat = line_str[0].strip().lower().replace(" ", "_")
+            raw_val = ":".join(line_str[1:]).strip()
+
+            if stat in power_stats.keys():
+                if stat == "time_stamp":
+                    stat_val = re.search(
+                        r"\d{2}/\d{2}/\d{4}\s-\s\d{2}:\d{2}:\d{2}", raw_val
+                    ).group(0)
```
DavidFair, line 128: When regex is involved, we should have line-by-line comments explaining what you are doing, as it can quickly suck up dev time

Ideally, I would push you to simplify with tools like `str.split()` then filter using lists instead of regex. It's more terse, but always way easier to maintain or change long term

cloud-energy-collection/iriscasttools/utils.py
```diff
@@ -0,0 +1,208 @@
...
+        "os_load_15: avg os loat last 15 minutes"
+    """
+    res = {x: "" for x in args}
+
+    stats = {"os_load_1": "", "os_load_5": "", "os_loads_15": ""}
+
+    try:
+        stats["os_load_1"], stats["os_load_5"], stats["os_load_15"] = os.getloadavg()
+    except OSError:
+        pass
```
DavidFair, line 159: These should warn to a logger if they fail

cloud-energy-collection/iriscasttools/utils.py
```diff
@@ -0,0 +1,208 @@
...
+            )
+        )
+
+        if stats["max_ram_kb"] and stats["used_ram_kb"]:
+            stats["ram_usage_percentage"] = round(
+                (stats["used_ram_kb"] / stats["max_ram_kb"]) * 100, 3
+            )
+
+    except (AssertionError, ValueError):
+        pass
```
DavidFair, line 205: Ditto

DavidFair (changes requested):

anish-mudaraddi:
can't figure out why the python 3.6 test is failing. Other than that it's ready to review

DavidFair (commented):
The pylint error is due to a slightly non-standard layout

If we move it to the following structure:
```
cloud-energy-collection/main.py
cloud-energy-collection/test/*
cloud-energy-collection/iriscasttools/utils.py
cloud-energy-collection/iriscasttools/*.py
```

Then we simply do `cd cloud-energy-collection && pylint --recursive .`  (notice test isn't included to add it twice)

I'm happy for you to call the top level dir to `cloud-energy-collection` to `iriscasttools`. It wouldn't be the only thing with a bespoke name :wink:

DavidFair:
> I'm going to change the structure of the files to this: ...

Just needs one tweak, `main.py` should sit along `requirements.txt`
The idea being, main imports `iriscasttools.utils` when it's running. In the future the implementation will all go into the subpackage whilst "users" simply use main from the outside.

anish-mudaraddi:
> Just needs one tweak, `main.py` should sit along `requirements.txt`

I added `__main__.py` so the users can call `python -m iriscastools <args>` 
I thought that made more sense :)

anish-mudaraddi:
> I prefer the second approach,

it seems better practice to keep tests outside the package 
https://doc.pytest.org/en/latest/explanation/goodpractices.html#tests-outside-application-code

DavidFair:
> > I prefer the second approach,
> 
> it seems better practice to keep tests outside the package https://doc.pytest.org/en/latest/explanation/goodpractices.html#tests-outside-application-code

I'll leave this up to your call, there's pros and cons

Outside:
Pros: You don't build them into the wheel. No chance someone would ever import something from test into the main application
Cons: Have to handle Python paths locally and setting up the CI. Some IDEs struggle

Inside:
Pros: Can run unit tests after installing your wheel if you want to test against a target platform.
Cons: Wheel size / above

Beyond that, it's a developer's call

anish-mudaraddi:
> I'll leave this up to your call, there's pros and cons
I'll leave the tests outside

I'm happy with this - I'll look into setting up wheels once this is merged

iriscasttools/iriscasttools/__main__.py
```diff
@@ -0,0 +1,5 @@
+import sys
```
DavidFair, line 1: This file should really be called `main` as the main entrypoint

iriscasttools/iriscasttools/__main__.py
```diff
@@ -0,0 +1,5 @@
...
+from iriscasttools.main import get_iriscast_stats, parse_args
+
+cmd_args = parse_args(sys.argv[1:])
+print(get_iriscast_stats(cmd_args.as_csv, cmd_args.include_header))
```
DavidFair, line 5: It would be nice to pre-configure the logger for the module like this:
https://github.com/stfc/SCD-OpenStack-Utils/blob/master/OpenStack-Rabbit-Consumer/entrypoint.py#L9

That way we can easily do something like
`LOG_LEVEL=DEBUG python3 -m iriscasttools` if something isn't working
...

iriscasttools/iriscasttools/utils.py
```diff
@@ -0,0 +1,224 @@
...
+        backoff: a factor to increase delay after every fail
+        retry_logger: logger to log failed attempts
+    """
+
+    def decorator(func):
+        def inner(*args, **kwargs):
+            res = None
+            for i in range(retries + 1):
+                try:
+                    res = func(*args, **kwargs)
```
DavidFair, line 38: Can we add debug logging for what args and kwargs we're about to try before
and something to debug log the return code

iriscasttools/iriscasttools/utils.py
```diff
@@ -0,0 +1,224 @@
...
+                    )
+                time.sleep(seconds)
+            return res
+
+        return inner
+
+    return decorator
+
+
+@retry(retry_on=(AssertionError,), retries=3, delay=3, backoff=2, retry_logger=logger)
```
DavidFair, line 62: Don't bother passing default values from above. Let's simplify to simply:

```
@retry(retry_on=(AssertionError,))
```

iriscasttools/iriscasttools/utils.py
```diff
@@ -0,0 +1,224 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def retry(
+    retry_on: (Exception,) = None,
+    retries: int = 3,
+    delay: int = 3,
+    backoff: int = 2,
+    retry_logger: logging.Logger = None,
```
DavidFair, line 21: We can capture the logger from the module scope here to remove boilerplate later on:

```
retry_logger: logging.Logger = logger
```

iriscasttools/iriscasttools/utils.py
```diff
@@ -0,0 +1,224 @@
...
+    """Run a bash command with given arguments and return output
+
+    Keyword Arguments
+        cmd_args: a string representing bash command to run
+    """
+    with subprocess.Popen(
+        cmd_args, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    ) as out:
+        # assert we did not find any errors
+        assert not out.stderr.read().decode()
```
DavidFair, line 73: Throw an exception instead of assert

Assert implies a programming error in Python
Exception implies a logical / runtime error, such as the OS not having the package

iriscasttools/iriscasttools/utils.py
```diff
@@ -0,0 +1,224 @@
...
+
+def check_ipmi_conn():
+    """Check if IPMItool exists and can be connected to
+
+    Checks if device exists at any of these locations /dev/ipmi0, /dev/ipmi/0 or /dev/ipmidev/0
+    which imply that ipmi-dcmi can be used to get power info
+    """
+    res = any(Path(f).exists() for f in ["/dev/ipmi0", "/dev/ipmi/0", "/dev/ipmidev/0"])
+    if not res:
+        logger.error("Failed to find ipmi device on host")
```
DavidFair, line 86: Ditto on these throwing if it doesn't make sense to continue throughout this file

iriscasttools/iriscasttools/utils.py
```diff
@@ -0,0 +1,224 @@
...
+        "time_stamp": current timestamp for reading taken
+        "power_measurement": whether power measurement is active on host
+    """
+    power_stats = {x: "" for x in args}
+
+    if not check_ipmi_conn():
+        return power_stats
+    res = ipmi_raw_power_query()
+
+    if res:
```
DavidFair, line 138: If `ipmi_raw_power_query` throws this whole block can be simplified by removing the if condition

iriscasttools/setup.py
```diff
@@ -0,0 +1,19 @@
...
+
+setup(
+    name="iriscasttools",
+    version=VERSION,
+    author="Anish Mudaraddi",
+    author_email="<anish.mudaraddi@stfc.ac.uk>",
+    description=DESCRIPTION,
+    long_description=LONG_DESCRIPTION,
+    packages=find_packages(),
+    python_requires=">=3",
```
DavidFair, line 16: Nit: >=3.6

iriscasttools/test/test_utils.py
```diff
@@ -0,0 +1,276 @@
...
+        expected_out_fp -- str, filepath to example expected values as csv
+    """
+    with open(test_fp, "r", encoding="UTF-8") as txt_file:
+        test_str = txt_file.read()
+
+    with open(expected_out_fp, "r", encoding="UTF-8") as csv_file:
+        csv_out = csv.DictReader(csv_file)
+        for row in csv_out:
+            expected_out = row
+            break
```
DavidFair, line 39: Nit: do this outside the with context

iriscasttools/iriscasttools/utils.py
```diff
@@ -0,0 +1,224 @@
...
+                try:
+                    res = func(*args, **kwargs)
+                    break
+                except retry_on:
+                    if i == retries:
+                        if retry_logger:
+                            retry_logger.error(
+                                "function failed and max retries %s exceeded", retries
+                            )
+                        return None
```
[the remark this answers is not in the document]
anish-mudaraddi: That's fair 
The reason I did this was so we didn't want miss collecting other metrics if one fails

DavidFair (approved):
LGTM

PR #50, opened 2023-03-21
Add script to benchmark GPUs

DavidFair:
Adds a script building onto the STFC SciML team's benchmarking tool. This runs the stemdl training steps across n GPUs to evaluate the FP.

It intentionally avoids setting up Tensorflow ...etc. to benchmark floating point perf only.


gpu_benchmark/benchmark.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+
```
meoflynn, line 2: Could we add a comment just to say which repo the git commit referenced is from?

meoflynn (commented):
Looks good to me, just one tiny change for the `benchmark.sh` file

.github/workflows/gpu_benchmark.yaml
```diff
@@ -0,0 +1,13 @@
+name: GPU Benchmark PR
+on: [push, pull_request]
```
anish-mudaraddi, line 2: you might want to change this to match CI/CD improvements PR
```suggestion
on:
  push:
    branches:
      - master
  pull_request:
    paths:
      - ".github/workflows/gpu_benchmark.yaml"
      - "gpu_benchmark/**"
```

anish-mudaraddi (commented):
Looks good, but a suggested change to match the other PR #51

anish-mudaraddi (approved):
looks good

PR #56, opened 2023-03-24
Use Hashcat and Phoronix to benchmark

DavidFair:
The previous implementation was IO bound on higher performance cards. Instead we switch to:

Phoronix Test Suite - This runs per card benchmarks and outputs the total run time

Hashcat - This acts as a stress test and captures power data

### DIFF
--- a/gpu_benchmark/benchmark.sh
+++ b/gpu_benchmark/benchmark.sh
@@ -2,85 +2,125 @@
 
 set -e
 
-# https://github.com/stfc-sciml/sciml-bench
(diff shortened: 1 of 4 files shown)


apdibbo:
You have a script to setup the gpu drivers but it's not mentioned in the readme

DavidFair:
> You have a script to setup the gpu drivers but it's not mentioned in the readme

This is called by the benchmark and power usage scripts if `nvidia-smi` fails (on the assumption it's a new install)

apdibbo:
Ok, cool

apdibbo (approved):

PR #59, opened 2023-04-14
Add in nulls for gpu fields for cpu flavors

apdibbo:
This means that vande will show and aggregate the CPU flavors


anish-mudaraddi (approved):
lgtm - although this file could use a re-write to use f-strings maybe

PR #62, opened 2023-05-02
Release Multiple RPM Versions

anish-mudaraddi:
separate out building rpms and wheel as two distinct steps - build multiple RPMs for different python versions 
- most notably python3.6 which is required for package to work on HVs

Previous PR #61 built RPM against python 3.11 - which doesn't work on HVs

### DIFF
--- a/.github/workflows/iriscast_package_build.yaml
+++ b/.github/workflows/iriscast_package_build.yaml
@@ -35,7 +35,7 @@ jobs:
         run: |
           cd iriscasttools && pylint --recursive yes test iriscasttools
 
-  build:
+  build-wheel:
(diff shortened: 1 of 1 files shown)


DavidFair:
Is this targetting SL7 then? AFAIK all of our HVs should be draining to reinstall or on RL8

anish-mudaraddi:
> Is this targetting SL7 then? AFAIK all of our HVs should be draining to reinstall or on RL8

Currently it builds noarch RPMs - so they can be used on any architecture - its what I used previously on the HVs for Iriscast work

DavidFair (approved):

PR #65, opened 2023-05-15
Change RPM package installation location

anish-mudaraddi:
makes it so the RPMs get installed to `/usr/lib/python{python-version.major}.{python-version.minor}//site-packages` instead of `/opt/hostedtoolcache/Python/3.6.8/x64/lib/python{python-version.major}.{python-version.minor}/site-packages/` 

e.g. 3.6 RPM `/usr/lib/python3.6/site-packages` instead of `/opt/hostedtoolcache/Python/3.6.8/x64/lib/python3.6/site-packages/`

### DIFF
--- a/.github/workflows/iriscast_package_build.yaml
+++ b/.github/workflows/iriscast_package_build.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ["3.x", "3.10", "3.6.8"]
+        python-version: ["3.x", "3.10", "3.6"]
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/iriscast_package_build.yaml
```diff
@@ -84,6 +84,8 @@ jobs:
       - name: Build RPM
         run: |
           cd iriscasttools
+          python_version=$(python -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')")
```
DavidFair, line 87: Shouldn't we use the matrix / var input from the CI rather than rely on Python echoing to us? Feels a bit like we're going to long way around

anish-mudaraddi: We have python '3.x' in matrix so we can't directly get the python version

DavidFair: In that case is it worth adding something like:
```
if: matrix.version != 3.x
```

Since we probably want to build wheels for specifically targetted major versions (hence 3.6 and 3.10)

PR #71, opened 2023-05-23
Fix wrong ip registration

DavidFair:
Fixes a few bugs from testing:

- `get_server` was returning all servers (which was v. slow) then passing back the first one each time
- Avoid trying to pull network data from a VM that doesn't exist
- Fix some ordering problems hidden by the `get_server` issue
- Check the Openstack Image is AQ by using `AQ_OS` as a litmus test, before trying to deserialise the metadata


OpenStack-Rabbit-Consumer/test/test_message_consumer.py
```diff
@@ -253,25 +246,94 @@ def test_consume_create_machine_hostnames_good_path(
...
     """
     rabbit_message.payload.metadata.machine_name = "AQ-HOST1"
 
-    with (
-        patch("rabbit_consumer.message_consumer.VmData") as data_patch,
-        patch("rabbit_consumer.message_consumer.openstack_api") as openstack,
-    ):
-        openstack.get_server_networks.return_value = openstack_address_list
-
+    with (patch("rabbit_consumer.message_consumer.VmData") as data_patch,):
```
meoflynn, line 255: Why are we doing a tuple here?

DavidFair: It was from when we patched more than one thing, I'll fix this now :+1:

meoflynn (approved):

PR #74, opened 2023-06-08
Use vm personality first

DavidFair:
Blocked by #72 

This PR updates the flow to use a VM's metadata first over the image metadata
Our custom flow for creating a Aq Managed VM (with bespoke personality etc) was broken by the rework, so this adds back that functionality


OpenStack-Rabbit-Consumer/rabbit_consumer/openstack_api.py
```diff
@@ -71,7 +71,7 @@ def get_server_networks(vm_data: VmData) -> List[OpenstackAddress]:
     return OpenstackAddress.get_internal_networks(server.addresses)
 
 
-def get_metadata(vm_data: VmData) -> dict:
+def get_server_metadata(vm_data: VmData) -> dict:
```
meoflynn, line 74: Is this method used anywhere else for the consumers and has the change in method name been updated elsewhere (in the consumers and the unit tests) where the method is called?

From what I can see this method isn't being currently used?

DavidFair: I've just double checked, it's called in `message_consumer`

meoflynn (approved):

PR #75, opened 2023-06-08
Fix double registration delete

DavidFair:
Blocked by #74 

The hostname implicitly will delete the network adapter, so we should shift this to an if-else rather than both always condition

This resolves an issue where the delete would fire twice and fail the second time


OpenStack-Rabbit-Consumer/rabbit_consumer/message_consumer.py
```diff
@@ -110,7 +122,7 @@ def check_machine_valid(rabbit_message: RabbitMessage) -> bool:
         )
         return False
 
-    if not is_aq_managed_image(rabbit_message):
+    if not is_aq_managed_image(vm_data):
         logger.debug("Ignoring non AQ Image: %s", rabbit_message)
```
meoflynn, line 126: Is `rabbit_message` still meant to be here if it's been replaced by `vm_data` in the line above?

meoflynn (approved):

PR #77, opened 2023-06-09
Add aq zombie finder

charlieDurnsfordHollands:
Adds a script that returns 2 lists, one of Cloud VMs that no longer exist and need to be removed from Aquilon, and another where the Cloud VMs that should be managed by Aquilon aren't.


.github/workflows/aq_zombie_finder.yaml
```diff
@@ -0,0 +1,34 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - "aq_zombie_finder/**"
+      - ".github/workflows/aq_zombie_finder.yaml"
+
+jobs:
+  test_and_lint:
```
anish-mudaraddi, line 13:
```suggestion
  test_with_unittest:
```

You aren't running a linter here. So this is just to make it clearer
If you want to run a linter you can add this to the end of the file:
```
    - name: Analyse with pylint 
      run: |
          pylint --recursive yes test aq_zombie_finder
```
you'll also need to put a `.pylintrc` file in `aq_zombie_finder` folder - you can copy the one in the iriscasttools folder 

I recommend using a linter - it will basically analyse your code and give you potential ways to improve your code

aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -0,0 +1,117 @@
+import paramiko
+from re import compile
+from getpass import getpass
+
+
+# Creates an SSH client through Paramiko with the users details and returns the client
+def create_client(host, user, password):
```
anish-mudaraddi, line 7: you've got some good comments here, but I would recommend writing docstrings instead of comments 

```suggestion
def create_client(host, user, password):
"""
Function to create an SSH client through Paramiko
    :param host: hostname to connect to (string)
    :param user: user to connect as (string)
    :param password: password for the user (string)
    :returns:  Configured Paramiko SSHClient object 
 """
```
A docstring is a comment that describes what the function does, what inputs it takes, and what it returns. I'd recommend changing your comments into docstrings

aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -0,0 +1,117 @@
...
+    if "No server with a name or ID of" in str(server_details):
+        # If the serial doesn't return a host, save its serial and IP as a potential zombie
+        aquilon_zombie_file.write(serial + " | " + aq_ip + "\n")
+
+
+def aq_zombie_finder():
+    # Prompt the user for their FedID, Password and Openstack IP to connect to
+    user = input("FedID: ")
+    password = getpass("Password: ")
+    openstack_ip = input("Openstack IP: ")
```
anish-mudaraddi, line 77: `aq_zombie_finder()` function is quite big - you can split this into multiple smaller helper functions - a good example would be getting user input

```
def get_user_input():
    user = input("FedID: ")
    password = getpass("Password: ")
    openstack_ip = input("Openstack IP: ")
    return user, password, openstack_ip
    
 and call that function in `aq_zombie_finder` here

File: aq_zombie_finder/aq_zombie_finder.py [hunk 4]
```diff
@@ -0,0 +1,117 @@
...
+
+    # Create a paramiko SSH client to the VM running Openstack and to Aquilon
+    openstack_client = create_client(openstack_ip, user, password)
+    aquilon_client = create_client("aquilon.gridpp.rl.ac.uk", user, password)
+
+    # Create a blank text documents for the output to be saved to
+    open("output\\openstack_zombie_list.txt", "w").close()
+    open("output\\aquilon_zombie_list.txt", "w").close()
```
anish-mudaraddi, line 85: when defining filepaths as string literals you should use `os.path.join`
```
openstack_zombie_filepath = os.path.join("output", "openstack_zombie_list.txt")
aquilon_zombie_filepath = os.path.join("output", "aquilon_zombie_list.txt")
```
You should probably also check if that file already exists
```
if os.path.exists(openstack_zombie_filepath):
    raise RuntimeError(f"{openstack_zombie_filepath} already exists"
```
The checks and creation can go in a separate function to - something like 
```
def create_output_files(openstack_zombie_filepath, aquilon_zombie_filepath):
    # checks and file creation here
```
This file doesn't need to return anything, but will raise an error if the file already exists

anish-mudaraddi, line 85: You should also have a way so the user of the script can specify the filepaths here - maybe as another input like your doing or using `argparse`

aq_zombie_finder/aq_zombie_finder.py
anish-mudaraddi, line 77: using `input` is fine, but this means that it's hard to automate python scripts.

Imagine we want to run this script to run every few days automatically - using `input` would force us to manually enter these credentials every time

I would recommend looking into `argparse` - this is a way were we can define inputs into the script when we run it - for example here we should be able to the script like so:

```
python3 aq_zombie_finder.py --user <some-user> --password <some-password> --openstack-ip <some-ip>
```

aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -0,0 +1,117 @@
...
+    openstack_zombie_file = open("output\\openstack_zombie_list.txt", "a")
+    aquilon_zombie_file = open("output\\aquilon_zombie_list.txt", "a")
```
anish-mudaraddi, line 87: using `open` and `close` to work with files is not recommended - it makes it harder to follow. use `with` instead
```
with open(openstack_zombie_filepath, 'a') as openstack_zombie_file, with open(aquilon_zombie_filepath, 'a') as aquilon_zombie_file:
    do things with file
```

aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -0,0 +1,117 @@
...
+    # Get a list of the IPs of AQ VMs
+    aq_ips = get_aq_ips(openstack_client)
+
+    # Run through each IP address found in the list of servers
+    for i, aq_ip in enumerate(aq_ips, start=1):
+        # Print the current progress through the list of IP addresses
+        print(i, "/", len(aq_ips))
+
+        # Check the ip of the Aquilon VM
+        aq_host = check_openstack_ip(aq_ip[0], aquilon_client, openstack_zombie_file)
```
anish-mudaraddi, line 98: Here you are opening the file first and then passing the opened file to the function to be used.  

I wouldn't recommend doing that - I would pass the filepath here instead and open the file in the function instead. 

It's just good programming etiquette - since other processes could be interacting with that same file - you want to minimize the amount of time your program interacts with that  file

aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # Check if an aquilon host exists
+        if aq_host:
+            # Check the serial number of the AQ host
+            check_aquilon_serial(
```
anish-mudaraddi, line 103: same here

anish-mudaraddi (changes requested):
Left a few comments on the actual code - overall its good but there's a few things you should change

charlieDurnsfordHollands:
> Left a few comments on the actual code - overall its good but there's a few things you should change
> 
> You can change your code and adds commits to this PR - when you're happy that you've addressed the issue you can mark it as resolved - feel free to continue the conversation for each change I requested if you have any questions.

I think I've addressed everything. I've decided to not include pylint as I was struggling to get it working and the black workflow already includes linting, so it felt redundant.

aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -0,0 +1,186 @@
...
+        "-o",
+        "--output",
+        metavar="OUTPUT",
+        help="Directory to create the output files in",
+    )
+
+    # Define the variables with the script arguments
+    args = parser.parse_args()
+    user = args.user
+    password = args.password
```
anish-mudaraddi, line 143: split this into a separate function - see #78 comment on the same thing

aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -0,0 +1,186 @@
...
+    aquilon_client = create_client("aquilon.gridpp.rl.ac.uk", user, password)
+
+    # Define the filepath for the output to be saved to
+    openstack_zombie_filepath = os.path.join(
+        output or "output", "openstack_zombie_list.txt"
+    )
+    aquilon_zombie_filepath = os.path.join(
+        output or "output", "aquilon_zombie_list.txt"
+    )
+    check_output_files(openstack_zombie_filepath, aquilon_zombie_filepath)
```
anish-mudaraddi, line 158: you can remove this function and just do the validation check with a for loop - see #78 for the same thing

aq_zombie_finder/test/test_aq_zombie_finder.py
```diff
@@ -0,0 +1,112 @@
...
+class AQZombieFinderTests(unittest.TestCase):
+    @patch("aq_zombie_finder.paramiko")
+    def test_create_client(self, mock_paramiko):
+        host = "test.aquilon.gridpp.rl.ac.uk"
+        user = "test.abc12345"
+        password = "test.A4gdfGs"
+
+        mock_client = MagicMock()
+        mock_paramiko.client.SSHClient.return_value = mock_client
+        mock_paramiko.client.set_missing_host_key_policy = MagicMock()
```
anish-mudaraddi, line 17:
```suggestion
```

anish-mudaraddi (changes requested):
left a few comments - similar changes to #78 - all in all good 
I recommend you use the coverage tool to see if you missed any unittests, 
https://www.pythontutorial.net/python-unit-testing/python-unittest-coverage/

you don't need 100% - 90-95% is a benchmark we use on other repos

anish-mudaraddi:
> black workflow already includes linting, so it felt redundant.

just to let you know - pylint is different from black - black formats your code to look the same, whilst pylint does more - it will analyse your code and point out things that don't look right 

Since this is a small script, you don't need to include it in github actions, but we'll want to in the future

I ran pylint on `aq_zombie_finder.py` 
```
************* Module aq_zombie_finder
aq_zombie_finder.py:37:0: C0301: Line too long (110/100) (line-too-long)
aq_zombie_finder.py:1:0: C0114: Missing module docstring (missing-module-docstring)
aq_zombie_finder.py:4:0: W0622: Redefining built-in 'compile' (redefined-builtin)
aq_zombie_finder.py:64:4: R1705: Unnecessary "else" after "return", remove the "else" and de-indent the code inside it (no-else-return)
aq_zombie_finder.py:66:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
aq_zombie_finder.py:95:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
aq_zombie_finder.py:111:0: C0116: Missing function or method docstring (missing-function-docstring)
aq_zombie_finder.py:4:0: C0411: standard import "from re import compile" should be placed before "import paramiko" (wrong-import-order)

------------------------------------------------------------------
Your code has been rated at 8.75/10
```

most of these are small things but its worth fixing them now

anish-mudaraddi (approved):
Okay this looks good - if you mark it ready for review we can merge it

PR #78, opened 2023-06-15
Add DNS Entry Checker

charlieDurnsfordHollands:
Adds a script that checks for missing and mismatching IPs and their DNS' in order to more easily identify and investigate them.

### DIFF
--- a/.github/workflows/dns_entry_checker.yaml
+++ b/.github/workflows/dns_entry_checker.yaml
@@ -1,0 +1,33 @@
+name: DNS Entry Checker Unittest
+
+on: 
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - "DNSEntryChecker/**"
+      - ".github/workflows/dns_entry_checker.yaml"
+
+jobs:
+  test_and_lint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9"]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "pip"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r DNSEntryChecker/requirements.txt
+      - name: Test with unittest
+        run: |
+            cd DNSEntryChecker
+            python3 -m unittest discover -s ./test -p "test_*.py"
--- a/.github/workflows/black.yaml
+++ b/.github/workflows/black.yaml
@@ -21,3 +21,8 @@ jobs:
         uses: psf/black@stable
         with:
           src: "iriscasttools/iriscasttools"
+          
+      - name: DNS Entry Checker 
+        uses: psf/black@stable
+        with:
+            src: "DNSEntryChecker"
--- a/DNSEntryChecker/dns_entry_checker.py
+++ b/DNSEntryChecker/dns_entry_checker.py
@@ -1,0 +1,114 @@
+import paramiko
+from re import compile
+from getpass import getpass
+from collections import defaultdict
+
+
+# Creates an SSH client through Paramiko with the users details and returns the client
+def create_client(host, user, password):
+    client = paramiko.client.SSHClient()
+    client.set_missing_host_key_policy(paramiko.AutoAddPolicy)
+    client.connect(hostname=host, username=user, password=password, timeout=60)
+    return client
+
+
+# Executes a command through a Paramiko client, waits for an exit status and outputs the response
+def ssh_command(client, command):
+    _stdin, _stdout, _stderr = client.exec_command(command, get_pty=True, timeout=60)
+    _stdout.channel.recv_exit_status()
+    return _stdout.readlines()
+
+
+# Checks if the IP address and DNS match
+def pair_ip_and_dns(dns_pair, order_check_dict, ip_rexp):
+    last_rexp = compile(r"([0-9]{1,3}\.[0-9]{1,3}(?!.))")
+
+    ips = ip_rexp.findall(dns_pair)
+    ips = ([ips[0][0], ips[1][0]])
+
+    order_values = last_rexp.findall(ips[1])[0].split(".")
+    order_check_dict[order_values[0]].append(order_values[1])
+    return ips
+
+
+# Checks if the DNS returns the matching IP
+def check_ip_dns_mismatch(ips, client, ip_rexp, backward_mismatch_file, forward_mismatch_file, backward_missing_file):
+    if ips[0].replace("-", ".") == ips[1]:
+        returned_dns = ssh_command(client, "dig -x {s} +short".format(s=ips[1]))
+        if returned_dns:
+            backward_ip = ip_rexp.findall(returned_dns[0])
+            if not backward_ip:
+                backward_missing_file.write(f"{ips[1]}\n")
+            elif not backward_ip[0][0] == ips[0]:
+                backward_mismatch_file.write(f"{ips[1]}\n")
+        else:
+            backward_missing_file.write(f"{ips[1]}\n")
+    else:
+        forward_mismatch_file.write(f"{ips[1]}\n")
+
+
+# Checks which IPs are missing from 2-254
+def check_missing_ips(key, gap_missing_file, i):
+    res = set([eval(j) for j in key[1]])
+    missing_values = [k for k in sorted(set(range(2, 255))) if k not in res]
+    for ip in missing_values:
+        gap_missing_file.write(f"172.16.{i + 100}.{ip}\n")
+
+
+def dns_entry_checker():
+    # Prompt the user for their FedID, Password and IP to connect to
+    user = input("FedID: ")
+    password = getpass("Password: ")
+    ip = input("IP: ")
+
+    # Create an SSH client with the credentials given
+    client = create_client(ip, user, password)
+
+    # Run a command on the VM to get all IP addresses with their domain names
+    dns_list = ssh_command(client, "dig @ns1.rl.ac.uk stfc.ac.uk  axfr | "
+                                   "grep -i nubes.stfc.ac.uk | "
+                                   "grep -v CNAME | "
+                                   "grep '172.16.105'")
+
+    # Create or clear, then open to append to, the output files
+    open("output\\forward_mismatch_list.txt", "w").close()
+    open("output\\backward_mismatch_list.txt", "w").close()
+    open("output\\forward_missing_list.txt", "w").close()
+    open("output\\backward_missing_list.txt", "w").close()
+    open("output\\gap_missing_list.txt", "w").close()
+    forward_mismatch_file = open("output\\forward_mismatch_list.txt", "a")
+    backward_mismatch_file = open("output\\backward_mismatch_list.txt", "a")
+    forward_missing_file = open("output\\forward_missing_list.txt", "a")
+    backward_missing_file = open("output\\backward_missing_list.txt", "a")
+    gap_missing_file = open("output\\gap_missing_list.txt", "a")
+
+    # Define the Regex used to identify IPs
+    ip_rexp = compile(r"(([0-9]{1,3}[.-]){3}[0-9]{1,3})")
+
+    # Create a dictionary to store IPs for checking gaps
+    order_check_dict = defaultdict(list)
+
+    # Check through all IP and DNS pairs received
+    for i, dns_pair in enumerate(dns_list):
+        print(i + 1, "/", len(dns_list))
+
+        ips = pair_ip_and_dns(dns_pair, order_check_dict, ip_rexp)
+
+        check_ip_dns_mismatch(ips, client, ip_rexp,
+                              backward_mismatch_file, forward_mismatch_file, backward_missing_file)
+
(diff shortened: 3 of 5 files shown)


.github/workflows/dns_entry_checker.yaml
```diff
@@ -0,0 +1,33 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - "dns_entry_checker/**"
+      - ".github/workflows/dns_entry_checker.yaml"
+
+jobs:
+  test_and_lint:
```
anish-mudaraddi, line 13: As mentioned in #77 you're not actually doing any linting 

```suggestion
  test_with_unittest:
```
You can add this to make github actions perform linting

```
    - name: Analyse with pylint 
      run: |
          pylint --recursive yes test dns_entry_checker
```
you'll also need to put a .pylintrc file in dns_entry_checker folder

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
+import paramiko
+from re import compile
+from getpass import getpass
+from collections import defaultdict
+
+
+# Creates an SSH client through Paramiko with the users details and returns the client
+def create_client(host, user, password):
```
anish-mudaraddi, line 8: Looks like you have some duplicate code here in #77 - this hints to me that perhaps we should combine your two scripts into a single package.

I'd say it's fine for now - but its worth looking at in the future.

anish-mudaraddi, line 8: Please Add Docstrings to these functions too - same as #77

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
...
+        if returned_dns:
+            backward_ip = ip_rexp.findall(returned_dns[0])
+            if not backward_ip:
+                backward_missing_file.write(f"{ips[1]}\n")
+            elif not backward_ip[0][0] == ips[0]:
+                backward_mismatch_file.write(f"{ips[1]}\n")
+        else:
+            backward_missing_file.write(f"{ips[1]}\n")
+    else:
+        forward_mismatch_file.write(f"{ips[1]}\n")
```
anish-mudaraddi, line 54: try to avoid nested ifs. 
As a rule of thumb - if you see something with 3 or more levels of indentation - it's a code smell - since its hard to follow logically 

```
    if not ips[0].replace("-", ".") == ips[1]:
        forward_mismatch_file.write(f"{ips[1]}\n")
        return 
    returned_dns = ssh_command(client, "dig -x {s} +short".format(s=ips[1]))
    if not returned_dns
        backward_missing_file.write(f"{ips[1]}\n")
        return
    backward_ip = ip_rexp.findall(returned_dns[0])
    if not backward_ip:
        backward_missing_file.write(f"{ips[1]}\n")
    elif not backward_ip[0][0] == ips[0]:
        backward_mismatch_file.write(f"{ips[1]}\n")
```

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
...
+    order_values = last_rexp.findall(ips[1])[0].split(".")
+    order_check_dict[order_values[0]].append(order_values[1])
+    return ips
+
+
+# Checks if the DNS returns the matching IP
+def check_ip_dns_mismatch(
+    ips,
+    client,
+    ip_rexp,
```
anish-mudaraddi, line 38: Since this variable is immutable (doesn't change)
you can set `ip_rexp` as a global variable at the top of the script
`IP_REXP = compile(r"(([0-9]{1,3}[.-]){3}[0-9]{1,3})")`

That way you don't have to use it here as a parameter

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
...
+        else:
+            backward_missing_file.write(f"{ips[1]}\n")
+    else:
+        forward_mismatch_file.write(f"{ips[1]}\n")
+
+
+# Checks which IPs are missing from 2-254
+def check_missing_ips(key, gap_missing_file, i):
+    res = set([eval(j) for j in key[1]])
+    missing_values = [k for k in sorted(set(range(2, 255))) if k not in res]
```
anish-mudaraddi, line 60: why are you doing sorted(set(range(2, 255)))?
surely this works fine
```suggestion
    missing_values = [k for k in range(2, 255) if k not in res]
```

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
...
+    for ip in missing_values:
+        gap_missing_file.write(f"172.16.{i + 100}.{ip}\n")
+
+
+def dns_entry_checker():
+    # Prompt the user for their FedID, Password and IP to connect to
+    user = input("FedID: ")
+    password = getpass("Password: ")
+    ip = input("IP: ")
```
anish-mudaraddi, line 69: see the comment on `argparse` in #77

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Create or clear, then open to append to, the output files
+    open("output\\forward_mismatch_list.txt", "w").close()
+    open("output\\backward_mismatch_list.txt", "w").close()
+    open("output\\backward_missing_list.txt", "w").close()
+    open("output\\gap_missing_list.txt", "w").close()
+    forward_mismatch_file = open("output\\forward_mismatch_list.txt", "a")
+    backward_mismatch_file = open("output\\backward_mismatch_list.txt", "a")
+    backward_missing_file = open("output\\backward_missing_list.txt", "a")
+    gap_missing_file = open("output\\gap_missing_list.txt", "a")
```
anish-mudaraddi, line 91: see the comment about not using open and close for files in #77

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Define the Regex used to identify IPs
+    ip_rexp = compile(r"(([0-9]{1,3}[.-]){3}[0-9]{1,3})")
+
+    # Create a dictionary to store IPs for checking gaps
+    order_check_dict = defaultdict(list)
```
anish-mudaraddi, line 97: why are you using defaultdict here and not just dict?

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,127 @@
...
+# Checks if the IP address and DNS match
+def pair_ip_and_dns(dns_pair, order_check_dict, ip_rexp):
+    last_rexp = compile(r"([0-9]{1,3}\.[0-9]{1,3}(?!.))")
+
+    ips = ip_rexp.findall(dns_pair)
+    ips = [ips[0][0], ips[1][0]]
+
+    order_values = last_rexp.findall(ips[1])[0].split(".")
+    order_check_dict[order_values[0]].append(order_values[1])
+    return ips
```
anish-mudaraddi, line 31: This function is doing 2 things - setting ips and populating order_values - this should be split into two separate functions which are called sequentially - one to get `ips` and another to populate `order_check_dict`

anish-mudaraddi (changes requested):

dns_entry_checker/dns_entry_checker.py
charlieDurnsfordHollands: Using a defaultdict means that if the list is appended to for a key that isn't in the dictionary yet, it automatically adds the key. Using a normal dict would work if line 57 is changed to:
```
    if order_values[0] not in order_check_dict.keys():
        order_check_dict[order_values[0]] = list(order_values[1])
    else:
        order_check_dict[order_values[0]].append(order_values[1])
```
instead of just:
```
order_check_dict[order_values[0]].append(order_values[1])
```
If the version using the dict is preferable then I'm fine with changing it though.

anish-mudaraddi: oh cool - did not know this - you can keep it as is.

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,223 @@
...
+        output or "output", "forward_mismatch_list.txt"
+    )
+    backward_mismatch_filepath = os.path.join(
+        output or "output", "backward_mismatch_list.txt"
+    )
+    backward_missing_filepath = os.path.join(
+        output or "output", "backward_missing_list.txt"
+    )
+    gap_missing_filepath = os.path.join(output or "output", "gap_missing_list.txt")
+    check_output_files(
```
anish-mudaraddi, line 189: you can rewrite check_output_files function to be more generic and call it multiple times for each file
something like 

```
for filepath in [forward_mismatch_filepath, backward_mismatch_filepath, backward_missing_filepath, gap_missing_filepath]:
    check_file(filepath)
```

dns_entry_checker/dns_entry_checker.py
```diff
@@ -0,0 +1,223 @@
...
+        metavar="OUTPUT",
+        help="Directory to create the output files in",
+    )
+
+    # Define the variables with the script arguments
+    args = parser.parse_args()
+    user = args.user
+    password = args.password
+    ip = args.ip
+    output = args.output
```
anish-mudaraddi, line 164: you can move this into another function - and return a dictionary of argument names : argument values 
```
def parse_args(inp_args):
    """
    Function to parse commandline args
    :param inp_args: a set of commandline args to parse (dict)
    :returns: A dictionary of parsed args 
    """
    ...code here..
    args = parser.parse_args(inp_args)
    return args
```
you can call the function like so
```
args = parse_args(sys.argv[1:])

File: dns_entry_checker/test/test_dns_entry_checker.py [hunk 12]
```diff
@@ -0,0 +1,193 @@
...
+class DNSEntryCheckerTests(unittest.TestCase):
+    @patch("dns_entry_checker.paramiko")
+    def test_create_client(self, mock_paramiko):
+        host = "test.aquilon.gridpp.rl.ac.uk"
+        user = "test.abc12345"
+        password = "test.A4gdfGs"
+
+        mock_client = MagicMock()
+        mock_paramiko.client.SSHClient.return_value = mock_client
+        mock_paramiko.client.set_missing_host_key_policy = MagicMock()
```
anish-mudaraddi, line 24: You don't need this line - since mock_paramiko will automatically set it to a magic mock
```suggestion
```

dns_entry_checker/test/test_dns_entry_checker.py
```diff
@@ -0,0 +1,193 @@
...
+        mock_paramiko.AutoAddPolicy = NonCallableMock()
+        mock_paramiko.connect = MagicMock()
+
+        client = create_client(host, user, password)
+
+        client.connect.assert_called_once_with(
```
anish-mudaraddi, line 30: Add this line to test that set_missing_host_key_policy function was called property
```suggestion
        client.set_missing_host_key_policy.assert_called_once_with(mock_paramiko.AutoAddPolicy)
        client.connect.assert_called_once_with(
```

dns_entry_checker/test/test_dns_entry_checker.py
```diff
@@ -0,0 +1,193 @@
...
+
+        mock_client.exec_command.assert_called_once_with(
+            command, get_pty=True, timeout=60
+        )
+        mock_stdout.channel.recv_exit_status.assert_called_once()
+        mock_stdout.readlines.assert_called_once()
+
+        self.assertEqual(command_output, mock_stdout.readlines.return_value)
+
+    def test_pair_ip_and_dns(self):
```
anish-mudaraddi, line 55: rename this test method so its more descriptive
```suggestion
    def test_find_ip_dns_pair_valid(self):
```

dns_entry_checker/test/test_dns_entry_checker.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+    @mock.patch("builtins.open")
+    def test_check_ip_dns_mismatch(self, name, returned_ips, expected_out, mock_file):
+        ips_dns_pair = returned_ips
+        client = MagicMock()
+        backward_mismatch_filepath = "test/backward/mismatch/filepath"
+        forward_mismatch_filepath = "test/forward/mismatch/filepath"
+        backward_missing_filepath = "test/backward/missing/filepath"
+
+        with patch("dns_entry_checker.ssh_command"):
```
anish-mudaraddi, line 75: why are you doing `with patch` here instead of `mock.patch`

dns_entry_checker/test/test_dns_entry_checker.py
```diff
@@ -0,0 +1,193 @@
...
+        check_missing_ips(key, gap_missing_filepath)
+
+        print(expected_out, mock_file.mock_calls)
+
+        if expected_out:
+            assert call("test/gap/missing/filepath", "a") in mock_file.mock_calls
+        else:
+            assert call("test/gap/missing/filepath", "a") not in mock_file.mock_calls
+
+
```
anish-mudaraddi, line 191: you're missing tests for check_output_files and populate_ip_dict

i'd recommend using `coverage` python tool to see if you covered all tests
https://www.pythontutorial.net/python-unit-testing/python-unittest-coverage/

anish-mudaraddi (changes requested):
left a few more comments

anish-mudaraddi (approved):
Okay this looks good - if you mark it ready for review we can merge it

PR #83, opened 2023-07-03
Added the vm creation script and unit tests

RJaliliSTFC:
Script to create vms on every hv in every aggregate

### DIFF
--- a/Openstack-VM-Creation-Script/vm_creation_v2.py
+++ b/Openstack-VM-Creation-Script/vm_creation_v2.py
@@ -10,11 +10,17 @@ conn = openstack.connect("openstack")
 
 
 def get_image():
+    """
+    returns images based on the hard coded string
+    """
     ubuntu_images = conn.image.images(name="ubuntu-focal-20.04-nogui")
     return ubuntu_images
 
 
 def select_single_image(images):
+    """
+    returns a single image based on the metadata status
+    """
     for image in images:
         if image.status == "active":
             useable_image = image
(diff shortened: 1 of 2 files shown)


Openstack-VM-Creation-Script/vm_creation_v2.py
```diff
@@ -0,0 +1,164 @@
...
+    for agg_name, flav in aggregate_flavor_list.items():
+        hvs = {agg_name: hv_list[agg_name]}
+        dict_of_aggregate_to_hv_objects.update(hvs)
+    return dict_of_aggregate_to_hv_objects
+
+
+def create_vms_on_hvs_with_space(hv_list, aggregate_flavor_list, image):
+    created_list = []
+    security_group_id = "771352ca-41b7-4fb6-addc-882aa0d1d628"
+    network_id = "fa2f5ebe-d0e0-4465-9637-e9461de443f1"
```
meoflynn, line 112: Are these meant to be hard coded into the method? 
If they are needed could you add a comment to each one about what security group and network goes with the ids?

Openstack-VM-Creation-Script/vm_creation_v2.py
```diff
@@ -0,0 +1,164 @@
...
+
+    current_image = select_single_image(image)
+
+    print(
+        create_vms_on_hvs_with_space(
+            dict_of_flavor_to_hypervisor_objects,
+            aggregate_to_smallest_flavor_dict,
+            current_image,
+        )
+    )
```
meoflynn, line 164: Could you add a comment about what this does?

meoflynn (changes requested):
Looks good so far, could you add docstrings to methods in the script and the unittests to describe what each method does?

Openstack-VM-Creation-Script/vm_creation_v2.py
RJaliliSTFC: Added a comment for their ID's, not been discussed if they should be a variable or not, they are also currently dev-cloud IDs which I will change as I don't think the script was meant to be run for the dev-cloud

DavidFair:
We should copy over / adapt one of the workflows to run pylint, black and the unit tests that we now have

DavidFair:
Can we close this since we've moved it to the st2 pack?

PR #86, opened 2023-07-06
Ensure output directory exists

charlieDurnsfordHollands:
GitHub doesn't include empty folders, so as such a check is required to ensure that the output folder exists, either by default or where specified.

### DIFF
--- a/aq_zombie_finder/aq_zombie_finder.py
+++ b/aq_zombie_finder/aq_zombie_finder.py
@@ -153,9 +153,7 @@ def aq_zombie_finder():
     openstack_zombie_filepath = os.path.join(
         output_location, "openstack_zombie_list.txt"
     )
-    aquilon_zombie_filepath = os.path.join(
-        output_location, "aquilon_zombie_list.txt"
-    )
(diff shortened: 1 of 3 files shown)


aq_zombie_finder/aq_zombie_finder.py
```diff
@@ -139,19 +140,20 @@ def aq_zombie_finder():
     user = args.user
     password = args.password
     openstack_ip = args.ip
-    output = args.output
+    output_location = args.output or "output"
```
meoflynn, line 143: Here, it "output" the default string if `args.output = None`?

charlieDurnsfordHollands: I've moved the default to be set in the arguements to make it more readable.

meoflynn (changes requested):
Looks good so far, just a couple small changes needed

DavidFair (approved):

DavidFair:
@charlieDurnsfordHollands Once Megan's comments are addressed I'd be happy to merge
Also at the moment, you have this marked as a draft? If you're ready for it to be merged can you `Ready for review`?

DavidFair (approved):

PR #88, opened 2023-07-13
CI + CD For Rally

DavidFair:
Adds CI and CD for rally in-line with our other source directories


apdibbo (approved):
Seems to still work so approving

PR #89, opened 2023-07-17
Fix issue where some failure types dont get reported

apdibbo:
Fix issue where some failure types dont get reported by the rally tests


anish-mudaraddi:
closing this and rewriting this code with tests in a follow up PR

PR #91, opened 2023-08-09
Ticket Word Cloud Generation Script

charlieDurnsfordHollands:
Create a script to generate a word cloud from the tickets in a queue. Requirements can be seen [here](https://stfc.atlassian.net/browse/CLDSCRM-366).


word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,254 @@
...
+    :param headers: A request Header (dict)
+    :param url: The URL to send the request (string)
+    :returns: A dictionary of JSON values
+    """
+    session = requests.session()
+    session.headers = headers
+    session.auth = auth
+
+    attempts = 5
+
```
anish-mudaraddi, line 94:
```suggestion
    response = None
```
response is referenced before assignment

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,254 @@
...
+    :param issue_filter: Dict of filters to check the issues against (dict)
+    :param auth: A HTTPBasicAuth object for authentication (HTTPBasicAuth)
+    :param headers: A request Header (dict)
+    :param host: The host used to create the URL to send the request (string)
+    :returns: A list with the contents of all valid issues
+    """
+    issues_contents = []
+    return_amount = 50
+    issues_amount = 0
+    while return_amount == 50:
```
anish-mudaraddi, line 126: If I understand your code correctly - you are running a while loop grabbing 50 at a time until its less than 50 so maybe change this to while not less than 50
```suggestion
    while not return_amount < 50:
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,254 @@
...
+                return issues_contents
+            if filter_issue(issue, issue_filter, issue_date):
+                issue_contents = issue.get("fields").get("summary")
+                if issue_contents:
+                    issues_contents.append(issue_contents)
+
+        return_amount = json_load.get("size")
+        issues_amount = issues_amount + return_amount
+
+    return issues_contents
```
anish-mudaraddi, line 147: While the above comment is about changing the while loop - I personally don't like having a while loop check a condition - I prefer to use `while True` and use `break` to explicitly exit a loop on a condition. 
This is just easier for people to understand - you continue the loop until some condition which causes the loop to break

```suggestion
    curr_marker = 0
    check_limit = 50
    while True:
        url = f"{host}/rest/servicedeskapi/servicedesk/6/queue/182/issue?start={curr_marker}"
        json_load = get_response_json(auth, headers, url)
        issues = json_load.get("values")
        issues_contents = []
        i = 0
        for i, issue in enumerate(issues, 1):
            issue_date = datetime.strptime(
                issue.get("fields").get("created")[:10], "%Y-%m-%d"
            )

            if issue_date < datetime.strptime(issue_filter.get("end_date"), "%Y-%m-%d"):
                return issues_contents

            if filter_issue(issue, issue_filter):
                issue_contents = issue.get("fields").get("summary")
                if issue_contents:
                    issues_contents.append(issue_contents)

        if i < check_limit:
            break
        curr_marker += json_load.get("size")
    return issues_contents
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,254 @@
...
+
+def filter_issue(issue, issue_filter, issue_date):
+    """
+    Function to check if an issue passes the set filters
+    :param issue: A dict of an issues contents (dict)
+    :param issue_filter: Dict of filters to check the issues against (dict)
+    :param issue_date: The date that the issue was created (string)
+    :returns: If the issue passes the filters
+    """
+    if issue.get("fields").get("assignee"):
```
anish-mudaraddi, line 158: this is fine but is there a reason you are using `get` instead of issue["fields"]["assignee"]?

Using `get` implies that you want to include a default if the key is not found - which you might want to do

if you want to return false if the key isn't found - you could rewrite it like so:

```
def filter_issue(issue, issue_filter, issue_date):
    """
    Function to check if an issue passes the set filters
    :param issue: A dict of an issues contents (dict)
    :param issue_filter: Dict of filters to check the issues against (dict)
    :param issue_date: The date that the issue was created (string)
    :returns: If the issue passes the filters
    """
    
    fields = issue.get("fields", None)
    if not fields:
        return False

    assignee = fields.get("assignee", None)
    if not assignee:
        return False

    issue_assigned = assignee.get("displayName", None)
    assign_check = issue_filter.get("assigned", None)
    if not issue_assigned or not assign_check or issue_assigned != assign_check:
        return False

    if issue_date > datetime.strptime(issue_filter.get("start_date"), "%Y-%m-%d"):
        return False
    return True
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def generate_word_cloud(issues_contents, issue_filter, word_cloud_output_location):
+    """
+    Function to generate and save a word cloud
+    :param issues_contents: The summary of every valid issue (list)
+    :param issue_filter: Dict of filters to check the issues against (dict)
+    :param word_cloud_output_location: The output location for the word cloud to be saved to
+    """
+    matches = re.findall(r"((\w+([.'](?![ \n']))*[-_]*)+)", issues_contents)
```
anish-mudaraddi, line 178: That's a scary regex :) 
could you add a comment which explains what it does?

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,254 @@
...
+        issues_contents = " ".join(list(list(zip(*matches))[0]))
+    issues_contents = filter_word_cloud(issue_filter, issues_contents)
+    word_cloud = WordCloud(
+        width=2000,
+        height=1000,
+        min_font_size=25,
+        max_words=10000,
+        background_color="white",
+        collocations=False,
+        regexp=r"\w*\S*",
```
anish-mudaraddi, line 189: worth providing these as kwargs for the generate word cloud?
you could rewrite the function like so:
```
generate_word_cloud(issues_contents, issue_filter, word_cloud_output_location, **kwargs):
    def generate_word_cloud(issues_contents, issue_filter, word_cloud_output_location, **kwargs):
    """
    Function to generate and save a word cloud
    :param issues_contents: The summary of every valid issue (list)
    :param issue_filter: Dict of filters to check the issues against (dict)
    :param word_cloud_output_location: The output location for the word cloud to be saved to
    :param kwargs: A set of kwargs to pass to WordCloud
        - width
        - height
        - min_font_size
        - max_words
    """
   ...
   
    word_cloud = WordCloud(
        width=kwargs["width"],
        height=kwargs["height"],
        min_font_size=kwargs["min_font_size"],
        max_words=kwargs["max_words"],
        background_color="white",
        collocations=False,
        regexp=r"\w*\S*",
    )
    ...
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,254 @@
...
+
+    return issues_contents
+
+
+def word_cloud_generator():
+    """
+    Function to take arguments, generate the output location and run the
+    functions to the data for the word cloud and generate it
+    """
+    args = parse_args(argv[1:])
```
anish-mudaraddi, line 225: this shouldn't be here - see below

anish-mudaraddi (approved):

.github/workflows/word_cloud_generator.yaml
```diff
@@ -0,0 +1,33 @@
...
+          cache: "pip"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r word_cloud_generator/requirements.txt
+      - name: Test with unittest
+        run: |
+            cd word_cloud_generator
+            python3 -m unittest discover -s ./test -p "test_*.py"
```
DavidFair, line 33: Why do we need the extra flags? Can we just name the file directly here instead since long-term we'll want to switch this to nose anyway

word_cloud_generator/readme.md
```diff
@@ -0,0 +1,29 @@
...
+The script takes ~10 seconds to complete on a month of tickets
+
+Unit tests exist which test the logic of the methods the script uses, and the tests should be run whenever changes are made to the code.
+
+## Requirements
+
+requests: 2.31.0
+parameterized: 0.9.0
+python-dateutil: 2.8.2
+wordcloud: 1.9.2
```
DavidFair, line 16: Can we drop this since it will rapidly go out of date whenever someone touches the requirements file

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
+import re
```
khalford, line 1: Pylint would fail this import order. Line 1 should go to line 13.

DavidFair (changes requested):
Some minor things to save us hassle in the future
@anish-mudaraddi if you want to make the changes I'd be happy to merge since it's us just removing duplicated info rather than changing

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
...
+
+@dataclass
+class IssuesFilter(DataClassDictMixin):
+    output: str
+    start_date: str
+    end_date: str
+    word_cloud: str
+    assigned: Optional[str] = None
+    filter_for: Optional[str] = None
+    filter_not: Optional[str] = None
```
khalford, line 26: This could be put in a separate file?

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+def from_user_inputs(**kwargs):
+    """
+    Take the inputs from a argparse and populate a IssuesFilter dataclass and return it
+    :param kwargs: a dictionary of argparse values
+    """
+
+    return IssuesFilter.from_dict(kwargs)
```
khalford, line 35: Do you need to use `.from_dict()` here?
I'm wondering if this would be better?
```suggestion
    return IssuesFilter(**kwargs)
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+def parse_args(inp_args):
+    """
+    Function to parse commandline args
+    :param inp_args: a set of commandline args to parse (dict)
+    :returns: A dictionary of parsed args
```
khalford, line 42:
```suggestion
def parse_args(inp_args: Dict):
    """
    Function to parse commandline args
    :param inp_args: a set of commandline args to parse
    :returns: A dictionary of parsed args
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
...
+        metavar="OUTPUT",
+        help="Directory to create the output files in",
+        default="output",
+    )
+    parser.add_argument(
+        "-s",
+        "--start_date",
+        metavar="START_DATE",
+        help="Date to get issues from",
+        default=datetime.now().strftime("%Y-%m-%d"),
```
khalford, line 73: For readability can this expression be put into a variable first?
```suggestion
        default=someVarContainingExpression,
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
...
+    )
+    parser.add_argument(
+        "-e",
+        "--end_date",
+        metavar="END_DATE",
+        help="Date to get issues to",
+        default=(datetime.now() - relativedelta(months=1)).strftime("%Y-%m-%d"),
```
khalford, line 80: Similar to above. 
```suggestion
        default=someVar),
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+def filter_issue(issue, issue_filter, issue_date):
+    """
+    Function to check if an issue passes the set filters
+    :param issue: A dict of an issues contents (dict)
+    :param issue_filter: Dict of filters to check the issues against (dict)
+    :param issue_date: The date that the issue was created (string)
+    :returns: If the issue passes the filters
+    """
```
khalford, line 187:
```suggestion
def filter_issue(issue: Dict, issue_filter: DIct, issue_date: str) -> bool:
    """
    Function to check if an issue passes the set filters
    :param issue: A dict of an issues contents 
    :param issue_filter: Dict of filters to check the issues against 
    :param issue_date: The date that the issue was created 
    :returns: If the issue passes the filters
    """
```

word_cloud_generator/word_cloud_generator.py
```diff
@@ -0,0 +1,307 @@
...
+    Function to take arguments, generate the output location and run the
+    functions to the data for the word cloud and generate it
+    """
+    args = parse_args(argv[1:])
+    host = "https://stfc.atlassian.net"
+    username = args.username
+    password = args.password
+
+    issue_filter = from_user_inputs(**args.__dict__)
```
khalford, line 274: I think you can use `vars()` here. That way you aren't using a private method.
```suggestion
    issue_filter = from_user_inputs(vars(args))
```

khalford (approved):

.github/workflows/word_cloud_generator.yaml
anish-mudaraddi: why are we switching this to nose and not pytest

anish-mudaraddi (approved):

PR #93, opened 2023-09-08
Prototype Vault rabbit consumers

gmatthews20:
This is a prototype for vault integrated machine provisioning with openstack.

When a create message is received a wrapped secret id is requested and stored in etcd. On a delete message all secret ids associated with the approle and VM uuid are revoked.


jacob-ward:
Closing until we have effort to dedicate to deploying this, can be re-opened

PR #94, opened 2023-10-04
Added csv_to_dict.py as the first step to uploading data to Netbox.

khalford:


Netbox_CSV_Read/CSV/csv_to_dict.py
```diff
@@ -0,0 +1,32 @@
+import pandas as pd
+
+
+class CSVtoDict:
+
```
meoflynn, line 5: Could we add a docstring to describe what this class does?

meoflynn (changes requested):

Netbox_CSV_Read/CSV/csv_to_dict.py
khalford: I've added a brief Docstring.

meoflynn (approved):

PR #99, opened 2023-10-06
Added the format_dict.py module that converts normal csv dictionaries to Netbox friendly dictionaries.

khalford:
Pynetbox friendly dictionaries

### DIFF
--- a/Netbox_CSV_Read/CSV/csv_utils.py
+++ b/Netbox_CSV_Read/CSV/csv_utils.py
@@ -1,4 +1,5 @@
 import pandas
+from typing import List, Dict
 
 
 class CsvUtils:
@@ -8,7 +9,7 @@ class CsvUtils:
     def __init__(self):
         self.pd = pandas
 
-    def csv_to_python(self, file_path: str) -> dict:
+    def csv_to_python(self, file_path: str) -> Dict:
         """
         This method reads data from csv files and writes them to a dictionary.
         :param file_path: The file path of the CSV file to be read from.
@@ -19,7 +20,7 @@ class CsvUtils:
         return dataframe
 
     @staticmethod
-    def separate_data(data: dict) -> list:
+    def separate_data(data: dict) -> List:
         """
         This method reduces Pandas CSV to Dict conversion to individual dictionaries.
         :param data: The data from the CSV file
--- a/Netbox_CSV_Read/CSV/csv_utils.py
+++ b/Netbox_CSV_Read/CSV/csv_utils.py
@@ -1,19 +1,21 @@
-import pandas as pd
+import pandas
 
 
 class CsvUtils:
     """
     This class provides methods to read data from CSV files and allow the data to be easily read and used elsewhere.
     """
-    @staticmethod
-    def csv_to_python(file_path: str) -> dict:
+    def __init__(self):
+        self.pd = pandas
+
+    def csv_to_python(self, file_path: str) -> dict:
         """
         This method reads data from csv files and writes them to a dictionary.
         :param file_path: The file path of the CSV file to be read from.
         :return: Returns the data from the csv as a dictionary.
         """
-        with pd.read_csv(file_path) as dataframe:
-            dataframe = dataframe.to_dict(orient="list")
+        dataframe = self.pd.read_csv(file_path)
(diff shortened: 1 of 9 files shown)


Netbox_CSV_Read/CSV/format_dict.py
```diff
@@ -0,0 +1,69 @@
+from typing import Optional
+from Enums.dcim_device_id import DeviceInfoID
+from Enums.dcim_device_no_id import DeviceInfoNoID
+from Netbox_Api.netbox_connect import NetboxConnect
+from operator import attrgetter
+
+
+class FormatDict(NetboxConnect):
```
meoflynn, line 8: Could you add a docstring to describe this class?

Netbox_CSV_Read/Enums/dcim_device_id.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
+
+
+class DeviceInfoID(Enum):
```
meoflynn, line 4: Could you add a docstring to describe this class?

Netbox_CSV_Read/Enums/dcim_device_no_id.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+
+class DeviceInfoNoID(Enum):
```
meoflynn, line 4: Could you add a docstring to this class?

meoflynn (changes requested):

Netbox_CSV_Read/CSV/csv_utils.py
```diff
@@ -1,19 +1,21 @@
-import pandas as pd
+import pandas
```
DavidFair, line 1: `import pandas as pd`

Netbox_CSV_Read/CSV/csv_utils.py
```diff
@@ -1,19 +1,21 @@
...
 
 
 class CsvUtils:
     """
     This class provides methods to read data from CSV files and allow the data to be easily read and used elsewhere.
     """
-    @staticmethod
-    def csv_to_python(file_path: str) -> dict:
+    def __init__(self):
+        self.pd = pandas
```
DavidFair, line 9: ...^ then we don't need this

Netbox_CSV_Read/CSV/format_dict.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        if not api:
+            self.netbox = NetboxConnect(url, token).api_object()
+        else:
+            self.netbox = api
+        self.dicts = dicts
+        self.enums_id = DeviceInfoID
+        self.enums_no_id = DeviceInfoNoID
+
+    def iterate_dicts(self) -> list:
```
DavidFair, line 26: This should be `List` and `Dict`, as `list != List` (it's one of those Python gotchas that was only fixed in later versions which we partially have access to)

Netbox_CSV_Read/CSV/format_dict.py
```diff
@@ -0,0 +1,69 @@
...
+        if attr_string == "DEVICE_TYPE":
+            value = value.get(slug=netbox_value).id
+        elif attr_string == "LOCATION":
+            if type(site_value) == int:
+                site_name = self.netbox.dcim.sites.get(site_value).name
+                site_slug = site_name.replace(" ", "-").lower()
+            value = value.get(name=netbox_value, site=site_slug)
+        else:
+            value = value.get(name=netbox_value).id
+        return value
```
DavidFair, line 69: Nit: Format dict should only format dicts

This should be in a different class, such as `NetboxData`

DavidFair (changes requested):

Netbox_CSV_Read/CSV/format_dict.py
khalford: If the return typing should be uppercase List and Dict, does this apply to argument typing? like:
def func(values: List):
    pass
or
def func(values: list):
    pass

Netbox_CSV_Read/CSV/format_dict.py
khalford: 👍

meoflynn (changes requested):

PR #104, opened 2023-10-16
Write tests for untested code.

khalford:
Cleaned up tests. Renamed files and refactored file structure. Moved some methods to more appropriate locations.


Pynetbox_Data_Uploader/lib/utils/csv_to_dict.py
```diff
@@ -1,29 +1,70 @@
...
+        :return: Returns the formatted dictionary
+        """
+        for key in dictionary:
+            netbox_id = self.get_id_from_key(key=key, dictionary=dictionary)
+            dictionary[key] = netbox_id
+        return dictionary
+
+    def get_id_from_key(self, key: str, dictionary: Dict) -> Union[str, int]:
+        """
+        This method calls the get_id method to
```
meoflynn, line 44: I think this docstring is half complete

Pynetbox_Data_Uploader/tests/test_csv_to_dict.py
```diff
@@ -0,0 +1,77 @@
...
+    with patch("utils.csv_to_dict.read_csv") as mock_read_csv:
+        res = instance.csv_to_python(file_path)
+    mock_read_csv.assert_called_once_with(file_path)
+    mock_read_csv.return_value.to_dict.assert_called_once_with(orient="list")
+    assert res == mock_read_csv.return_value.to_dict.return_value
+
+
+def test_separate_data(instance):
+    """
+    This test ensures that the dictionaries from panda formatted into row by row dictionaries.
```
meoflynn, line 25: `pandas`

Pynetbox_Data_Uploader/tests/test_netbox_check.py
```diff
@@ -1,18 +1,12 @@
...
-def instance_api_fixture():
-    return MagicMock()
-
-
 @pytest.fixture(name="instance")
-def instance_fixture(api_mock):
-    url = "not real url"
-    token = "not real token"
-    return NetboxExistence(url, token, api_mock)
+def instance_fixture():
```
meoflynn, line 7: Could we add a docstring to this method?

Pynetbox_Data_Uploader/tests/test_netbox_create.py
```diff
@@ -1,18 +1,12 @@
...
-def instance_api_fixture():
-    return MagicMock()
-
-
 @pytest.fixture(name="instance")
-def instance_fixture(api_mock):
-    url = "not real url"
-    token = "not real token"
-    return NetboxDCIM(url, token, api_mock)
+def instance_fixture():
```
meoflynn, line 7: Could we add a docstring to this method?

Pynetbox_Data_Uploader/tests/test_netbox_data.py
```diff
@@ -0,0 +1,9 @@
+from unittest.mock import NonCallableMock
+from netbox_api.netbox_data import NetboxGetID
+import pytest
+
+
+@pytest.fixture(name="instance")
+def instance_fixture():
```
meoflynn, line 7: Could we add a docstring to describe this fixture method?

meoflynn (changes requested):

Pynetbox_Data_Uploader/.pylintrc
```diff
@@ -10,5 +10,6 @@ max-line-length=120
 # R0801: Similar lines - Imports and method signatures will flag this, such as forwarding action args
 # C0116: Missing method docstring - Adds too much noise
 # E0401: Ignore import errors as imports work
+# F0010: Ignore __init__.py files
 
-disable=C0114,C0115,E0401,W0511,R0801,C0116,E0401
```
DavidFair, line 14: We should drop this extra disable and add `__init__` files to the modules that are missing them

Pynetbox_Data_Uploader/lib/netbox_api/netbox_check.py
```diff
@@ -1,17 +1,13 @@
...
     """
     This class contains methods that check if an object exists in Netbox.
     """
 
-    def __init__(self, url: str, token: str, api: Optional = None):
-        if not api:
-            self.netbox = NetboxConnect(url, token).api_object()
-        else:
-            self.netbox = api
+    def __init__(self, netbox: Optional = None):
```
DavidFair, line 9: Can we type `NetBox` e.g.:
```
def __init__(self, netbox: Netbox = None)
```

Additionally you've marked it as optional, but if a user passes `None` we don't handle it below?

Pynetbox_Data_Uploader/lib/netbox_api/netbox_create.py
```diff
@@ -1,19 +1,15 @@
...
     This class contains methods that will interact create objects in Netbox.
     """
 
-    def __init__(self, url: str, token: str, api: Optional = None):
-        if not api:
-            self.netbox = NetboxConnect(url, token).api_object()
-        else:
-            self.netbox = api
+    def __init__(self, netbox):
+        self.netbox = netbox
```
DavidFair, line 10: Ditto

Pynetbox_Data_Uploader/lib/netbox_api/netbox_data.py
```diff
@@ -1,30 +1,21 @@
...
-        Also, it allows dependency injection testing.
-        :param url: Netbox website URL.
-        :param token: Netbox authentication token.
+        This method allows the Netbox Api Object and Enums to be accessible within the class.
         """
-        if not api:
-            self.netbox = NetboxConnect(url, token).api_object()
-        else:
-            self.netbox = api
+        self.netbox = netbox
```
DavidFair, line 17: ^

Pynetbox_Data_Uploader/lib/utils/csv_to_dict.py
```diff
@@ -1,29 +1,75 @@
...
+    This class takes dictionaries with string values and changes those to ID values from Netbox.
     """
 
+    def __init__(self, netbox: Optional = None):
+        """
+        This method initialises the class with the following parameters.
+        Also, it allows dependency injection testing.
+        """
+        self.enums_no_id = DeviceInfoNoID
+        self.netbox = netbox
```
DavidFair, line 18: Format dict shouldn't do anything which requires the netbox API....

Pynetbox_Data_Uploader/lib/utils/csv_to_dict.py
```diff
@@ -1,29 +1,75 @@
...
+        :param key: The attribute to look for.
+        :param dictionary: The device dictionary being referenced.
+        :return: If an ID was needed and found it returns the ID. If an ID was not needed it returns the original value.
+        """
+        if key not in list(self.enums_no_id.__members__):
+            value = NetboxGetID(self.netbox).get_id(
+                attr_string=key,
+                netbox_value=dictionary[key],
+                site_value=dictionary["site"],
+            )
```
DavidFair, line 54: Can we do this in something like the `NetBox` class?

Pynetbox_Data_Uploader/lib/utils/csv_to_dict.py
DavidFair, line 1: This file should be called `format_dict` to match the name of the class inside

DavidFair (changes requested):

Pynetbox_Data_Uploader/.pylintrc
khalford: Yep, didn't realise I needed those. Fixed a while ago, never removed the disable.

Pynetbox_Data_Uploader/lib/utils/csv_to_dict.py
khalford: It doesn't do anything with Netbox. It just passes the api object through to the netbox_data.py . This allows you to initialise the api once at the start of a script and pass the api around from there, Rather than moving the url and tokens around and initialising a new api object for every method/class.

khalford:
Pylint is going to fail on this branch due to the .pylintrc edit. I have already fixed these issues in another branch which I will merge once this branch merges.

Pynetbox_Data_Uploader/lib/utils/csv_to_dict.py
DavidFair: We need to add a file called `__init__.py` to utils to resolve the import error, since this is a module but we're missing the module "marker file"

khalford: utils does an `__init__.py` already in the directory.

DavidFair:
Sorry I sent you down the wrong track:

```
 tests/test_csv_to_dict.py:2: in <module>
    from utils.csv_to_dict import FormatDict
 ```

The test file name (csv_to_dict.py) and the import needs fixing for this case, to match your rename to `format_dict.py`

Once that's done I'm happy to pair program the pylint error, as suppressing it means it's not actually checking that file

DavidFair (approved):

meoflynn (approved):

PR #107, opened 2023-10-26
Add logging to note were clearing existing records

DavidFair:
Adds some extra logging, to just say why we're removing a VM during a creation as it might not be obvious


meoflynn (approved):

anish-mudaraddi (approved):

anish-mudaraddi (changes requested):
looks like your pylint is failing

anish-mudaraddi (approved):

PR #111, opened 2023-10-30
MAINT: Changed default rack height to 0.

khalford:
Simple change after discussing with others about physical rack space in R89 and R26. Rack height should be 0 as some "devices" such as PDU's don't take up any rack space. Whereas a HV does.


Koldstream (approved):

Pynetbox_Data_Uploader/lib/netbox_api/netbox_create.py
dev-0pz, line 1: for line 9

def __init__ requires a docstring e.g. " initialise for x netbox object"

dev-0pz (changes requested):

Pynetbox_Data_Uploader/lib/netbox_api/netbox_create.py
```diff
@@ -19,14 +19,14 @@ def create_device(self, data: Union[Dict, List]) -> bool:
...
-        self, model: str, slug: str, manufacturer: str, u_height: int = 1
+        self, model: str, slug: str, manufacturer: str, u_height: int = 0
     ) -> bool:
         """
         This method creates a new device type in Netbox.
         :param model: The model name of the device.
         :param slug: The URL friendly version of the model name.
         :param manufacturer: The manufacturer of the device.
-        :param u_height: This it the height of the device in the rack. Default 1.
+        :param u_height: This it the height of the device in the rack. Default 0.
```
anish-mudaraddi, line 29: typo?
```suggestion
        :param u_height: This is the height of the device in the rack. Default 0.
```

PR #112, opened 2023-10-30
Change default rack height to 0

khalford:
Simple change after discussing with others about physical rack space in R89 and R26. Rack height should be 0 as some "devices" such as PDU's don't take up any rack space. Whereas a HV does.

Fixes issues with CI/CD


dev-0pz (approved):

meoflynn (approved):
Looks good to me.

It looks like `requirements.txt` is still being seen by GitHub as a binary file, I think if it's just how GitHub is showing it in the PR I think it would be fine to merge

PR #114, opened 2023-10-30
Written missing test and renamed a file.

khalford:
Written a missing test for format_dict. Renamed netbox_data to match the class name.


anish-mudaraddi (approved):
LGTM - you just added 3 tests it seems - weird that github says you've got 600 odd changes

PR #115, opened 2023-10-30
Written tests for get_id_from_key method.

khalford:
Added tests for the above method and changed some enums stuff in the code. Removed attributes from netbox_get_id class as these enums will not be changed therefore do not need to be assigned.


anish-mudaraddi (approved):
LGTM - `get_id` should be split up in future

Koldstream (approved):

PR #116, opened 2023-11-06
Dictionaries to dataclass

khalford:
Created the device dataclass. Created 2 methods to take the csv file and create a dataclass object for each class.

### DIFF
--- a/pynetbox_data_uploader/.pylintrc
+++ b/pynetbox_data_uploader/.pylintrc
@@ -5,6 +5,5 @@ max-line-length=120
 
 # Disable various warnings:
 # C0114: Missing module string - we don't need module strings for the small repo
-# W0511: TODOs they're well....to do later
 
-disable=C0114,W0511
+disable=C0114
--- a/pynetbox_data_uploader/lib/user_methods/csv_to_netbox.py
+++ b/pynetbox_data_uploader/lib/user_methods/csv_to_netbox.py
@@ -1,4 +1,5 @@
 from typing import List
+from pathlib import Path
 import argparse
 from lib.netbox_api.netbox_create import NetboxCreate
 from lib.netbox_api.netbox_connect import NetboxConnect
@@ -35,6 +36,10 @@ class CsvToNetbox:
         :return: Returns a list of devices
         """
         print("Reading CSV...")
+        try:
+            Path(file_path).resolve()
+        except (OSError, RuntimeError):
+            raise Exception("The given path is not valid.")
         device_data = self.format_dict.csv_to_python(file_path)
         device_list = self.format_dict.separate_data(device_data)
         print("Read CSV.")
--- a/pynetbox_data_uploader/lib/user_methods/csv_to_netbox.py
+++ b/pynetbox_data_uploader/lib/user_methods/csv_to_netbox.py
@@ -14,6 +14,7 @@ class CsvToNetbox:
     """
     This class contains organised methods in the 4 step proccess of reading csv's to then uploading to Netbox.
     """
+
     def __init__(self, url: str, token: str):
         """
         This initialises the class with the following parameters.
--- a/pynetbox_data_uploader/lib/user_methods/csv_to_netbox.py
+++ b/pynetbox_data_uploader/lib/user_methods/csv_to_netbox.py
@@ -1,5 +1,5 @@
 from typing import List
-from pathlib import Path
+import pathlib
 import argparse
 from lib.netbox_api.netbox_create import NetboxCreate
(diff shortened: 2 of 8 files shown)


Protabs (approved):

pynetbox_data_uploader/lib/utils/csv_to_dataclass.py
```diff
@@ -0,0 +1,41 @@
...
+            status=row["status"],
+            site=row["site"],
+            location=row["location"],
+            rack=row["rack"],
+            face=row["face"],
+            airflow=row["airflow"],
+            position=row["position"],
+            name=row["name"],
+            serial=row["serial"],
+        )
```
anish-mudaraddi, line 39: This is fine - but you could create a method called `from_dict` in the dataclass to make this a bit tidier 
```
from dataclasses import dataclass, fields

from_dict(device_dict: Dict):
        field_set = {field.name for field in fields(Device) if field.init}
        filtered_arg_dict = {
            key: value for key, value in device_dict.items() if key in field_set
        }
        return Device(**filtered_arg_dict)
```

pynetbox_data_uploader/lib/utils/dataclass_data.py
```diff
@@ -0,0 +1,23 @@
...
+    device_type: str
+    status: str
+    site: str
+    location: str
+    rack: str
+    face: str
+    airflow: str
+    position: str
+    name: str
+    serial: str
```
anish-mudaraddi, line 23: Are any of these attribute optional? Can we set some of these to a default value?

pynetbox_data_uploader/lib/utils/csv_to_dataclass.py
```diff
@@ -0,0 +1,41 @@
+import csv
+from typing import List
+from lib.utils.dataclass_data import Device
+
+
+def open_file(file_path: str) -> csv.DictReader:
```
anish-mudaraddi, line 6: Whenever you're using file paths - you should use Pathlib to validate the filepath before you use it 

https://docs.python.org/3/library/pathlib.html

anish-mudaraddi (changes requested):
looks good - I made a few points - much of it can be ignored - just something to think about

pynetbox_data_uploader/.pylintrc
```diff
@@ -5,6 +5,5 @@ max-line-length=120
 
 # Disable various warnings:
 # C0114: Missing module string - we don't need module strings for the small repo
-# W0511: TODOs they're well....to do later
 
-disable=C0114,W0511
```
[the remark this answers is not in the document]
khalford: A David F nitpick. It hasn't been used anywhere in the code, so doesn't really need to be there.

pynetbox_data_uploader/lib/utils/dataclass_data.py
khalford: Some are optional, none have a default value.

pynetbox_data_uploader/lib/utils/csv_to_dataclass.py
khalford: Will consider that in future PR.

pynetbox_data_uploader/lib/utils/dataclass_data.py
anish-mudaraddi: if some are optional the default value should be None - you'll get an error if the user doesn't provide these. Set these values to None in the dataclass

pynetbox_data_uploader/lib/utils/dataclass_data.py
```diff
@@ -16,8 +16,8 @@ class Device:
     site: str
     location: str
     rack: str
-    face: str
-    airflow: str
+    face: Optional[str]
+    airflow: Optional[str]
```
anish-mudaraddi, line 20: if these are optional then you need to provide a default value of None

pynetbox_data_uploader/lib/user_methods/csv_to_netbox.py
```diff
@@ -35,6 +36,10 @@ def read_csv(self, file_path) -> List:
         :return: Returns a list of devices
         """
         print("Reading CSV...")
+        try:
+            Path(file_path).resolve()
+        except (OSError, RuntimeError):
+            raise Exception("The given path is not valid.")
```
anish-mudaraddi, line 42: did you add a test case for this? You can mock Path

Protabs (approved):

anish-mudaraddi (approved):

PR #122, opened 2023-11-20
User methods 2

khalford:
These changes are to hopefully make the code much more reusable and segmented. 
The read factory method allows files to be read in any format so long as an appropriate reading function has been written.
The validate factory method allows any data field in the device dataclass to be check in netbox so long as an appropriate function has been written to do that.
The argparse script now has subcommands that a user to display different help messages and show different command usages.
There are currently two user scripts: `upload_devices_to_netbox.py` and `validate_data_fields_in_netbox.py`. Both of these scripts should be very simple and only differ by calling each factory method with different arguments. 
The code is not finished yet and tests are not written/ do not work. Hence why this is a draft.
I have not made any changes to the `netbox_get_id` parts of the project yet as it's a little more complicated but would like to have that method in a sort of factory method like `read` and `validate`.
Docstrings should be very explanative and if they are not please say so!


pynetbox_query/pynetboxquery/utils/read_data.py
```diff
@@ -0,0 +1,96 @@
...
+# Disabling this pylint warning as it is unnecessary.
+# pylint: disable = R0903
+class ReadData(_ReadMethods):
+    """
+    This class contains methods to read data from different file types into a list of dictionaries.
+    Only the "read" method should be called by the user as it can differentiate between different file types.
+    The rest of the methods read data from the file into a list of dictionaries.
+    """
+
+    def read(self, file_path: str, **kwargs) -> List[Device]:
```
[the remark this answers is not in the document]
anish-mudaraddi: incidentally - this removes the need for `validate_kwargs`/ `check_filepath`  functions. Since that should be handled by the specific `Read` method

i.e. 

```
class ReadCSV(ReadFile):
    def __init__(**kwargs):
        self._validate(kwargs)
    
    def _validate(params: Dict):
        if "sheet_name" not in kwargs:
            raise error
    
    def read(file_path: str):
        self._check_file_path(file_path) # inherited from ReadFile Abstract Class
        with open(file_path, mode="r", encoding="UTF-8") as file:
            csv_reader = DictReader(file)
            dictionary_list = list(csv_reader)
        return dictionary_list

**Review by anish-mudaraddi — CHANGES_REQUESTED** [review]
Good progress - A couple of things namely about favoring inheritance over composition

File: pynetbox_query/pynetboxquery/utils/read_utils/read_abc.py [hunk 2]
```diff
@@ -4,20 +4,32 @@
...
-
     @abstractmethod
     def read(self) -> List[Dict]:
-        """"""
+        """
+        This method reads the contents of a file into a list od Device Dataclasses.
+        """
+
+    @staticmethod
+    def _validate(kwargs):
```
anish-mudaraddi, line 29: this should also be abstract?

pynetbox_query/pynetboxquery/user_methods/upload_devices_to_netbox.py
```diff
@@ -5,66 +5,55 @@
...
-def main():
-    """
-    This function collects the arguments and calls the upload function.
-    """
-    kwargs = _collect_args()
-    upload_devices_to_netbox(**kwargs)
+from pynetboxquery.user_methods.abstract_user_method import AbstractUserMethod
+
+
+class Main(AbstractUserMethod):
```
anish-mudaraddi, line 11: probably rename this to `UploadDevicesToNetbox`

pynetbox_query/pynetboxquery/user_methods/validate_data_fields_in_netbox.py
```diff
@@ -2,60 +2,50 @@
...
-def main():
-    """
-    This function collects the arguments and calls the validate function.
-    """
-    kwargs = _collect_args()
-    validate_data_fields_in_netbox(**kwargs)
+from pynetboxquery.user_methods.abstract_user_method import AbstractUserMethod
+
+
+class Main(AbstractUserMethod):
```
anish-mudaraddi, line 8: rename this to `ValidateFieldsInNetbox` or something similar

anish-mudaraddi (changes requested):
looks good - And the latest changes are great - good use of Abstract Classes - you've picked that up really well!

Id recommend renaming the `Main` classes so they're the same as the file name

khalford:
WooHoo!

anish-mudaraddi (commented):
Your tests look okay

maybe think about putting your tests in folders - so its the same as the directory structure of your actual code

anish-mudaraddi (dismissed):

anish-mudaraddi (approved):

dev-0pz (approved):
LGTM

PR #125, opened 2023-12-06
Word cloud generator 2

khalford:
Made changes from old PR: https://github.com/stfc/SCD-OpenStack-Utils/pull/91


anish-mudaraddi (approved):
This really needs improvement - particularly the tests need to be changed to pytest - but it works now so I'm not going to make you rewrite this :)

PR #128, opened 2024-02-02
ENH: Add monitoring script to collect information about VMs in different states

meoflynn:
Adds a new script for collecting information about VMs in different states with unit tests


MonitoringTools/usr/local/bin/collect_vm_stats.py
```diff
@@ -0,0 +1,121 @@
...
+    server_obj = conn.compute.servers(
+        details=False, all_projects=True, limit=10000, status="SHUTOFF"
+    )
+    # get number of items in generator object
+    instance_list = list(server_obj)
+    instance_shutoff = len(instance_list)
+    return instance_shutoff
+
+
+def collect_stats(cloud, prod):
```
anish-mudaraddi, line 82: NIT: make name clearer
```suggestion
def get_all_server_statuses(cloud, prod):
```

MonitoringTools/usr/local/bin/collect_vm_stats.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    Collects the stats for vms and returns a dict
+    :param cloud: OpenStack Cloud connection
+    :param prod: Boolean to determine whether Prod or Dev Cloud used
+    :return: A comma separated string containing VM states.
+    """
+    # raise error if cloud connection not given
+    if not cloud:
+        raise ValueError("An OpenStack Connection is required")
```
anish-mudaraddi, line 91: I would suggest passing the cloud `instance` ("prod" or "dev") as a parameter to this function and try to create the cloud connection object here instead


refactor the function like so. 
```python
def get_all_server_statuses(instance: str) -> str:
    ...
    conn = openstack.connect(instance)
    ... 
    return f"VMStats,instance={instance.capitalize()} totalVM={total_vms}i,activeVM={active_vms}i,buildVM={build_vms}i,errorVM={error_vms}i,shutoffVM={shutoff_vms}i"
```

this ties in with how you can incorporate the helper functions I wrote - which will post the data to influxdb

MonitoringTools/usr/local/bin/collect_vm_stats.py
```diff
@@ -0,0 +1,121 @@
...
+        cloud_env = "PreProd"
+
+    vm_stats = f"VMStats, instance={cloud_env} totalVM={total_vms}i,activeVM={active_vms}i,buildVM={build_vms}i,errorVM={error_vms}i,shutoffVM={shutoff_vms}i"
+
+    return vm_stats
+
+
+if __name__ == "__main__":
+    cloud_conn = openstack.connect(cloud="openstack")
+    print(collect_stats(cloud_conn, prod=False))
```
anish-mudaraddi, line 121: you'll want to change this. 

create a main function like so 
```python
def main(user_args: List):
   influxdb_args = parse_args(user_args, description="Get All VM Statuses")
   run_scrape(influxdb_args, get_all_server_statuses)

if __name__ == "__main__":
    main(sys.argv[1:])

```
This way we can run the script like so: 
`python3 collect_vm_stats.py /path/to/influxdb.conf` 

and it will automatically post to influx.

anish-mudaraddi (changes requested):
I haven't looked through your tests yet - but I've made comments on how you can use the stuff I already wrote to automatically post to influxdb

anish-mudaraddi:
Also you'll want to move your test file into the tests folder `/MonitoringScripts/tests` so the CI/CD will run them

khalford (changes requested):
Looks good. Just a few comments

anish-mudaraddi:
fixed a bug and added tests for prod. 
I'm not going for full test coverage for `run_server_query` because this functionality is replicated with the query library. 
and we will eventually move these scripts to stackstorm or get them using the query library, so we won't need to bugfix pagination problems

khalford (approved):

anish-mudaraddi (approved):

PR #130, opened 2024-02-09
BUG: fix various bugs with influxdb monitoring scripts

anish-mudaraddi:
slottifier now only selects hvs with matching storagetype and hosttype to calculate slots on

fix data string formatting for `service_status_to_influx.py` script


khalford (approved):
LGTM

DavidFair (approved):

PR #131, opened 2024-02-12
BUG: fix bugs with service-status-to-influx script

anish-mudaraddi:
update script to latest version on prod
fix not adding statustext to hv when nova-compute service found


Protabs (approved):

khalford (approved):
LGTM

PR #132, opened 2024-02-26
AQ network bug

khalford:
VMs not created on the Internal network are binned off by Rabbit. This fix allows Rabbit to check if they are also in the services subnet.


OpenStack-Rabbit-Consumer/rabbit_consumer/openstack_api.py
```diff
@@ -71,8 +71,10 @@ def get_server_networks(vm_data: VmData) -> List[OpenstackAddress]:
...
     """
     server = get_server_details(vm_data)
-    if "Internal" not in server.addresses:
-        logger.warning("No internal network found for server %s", server.name)
+    if "Internal" not in server.addresses and "Services" not in server.addresses:
+        logger.warning(
+            "No internal or services network found for server %s", server.name
+        )
         return []
     return OpenstackAddress.get_internal_networks(server.addresses)
```
meoflynn, line 79: We may need another method (like `get_service_networks`) to handle a server address that has a Services network

meoflynn (changes requested):

OpenStack-Rabbit-Consumer/tests/test_openstack_api.py
```diff
@@ -125,7 +123,21 @@ def test_get_server_networks(address, server_details, vm_data):
...
+    address.get_services_networks.assert_called_once_with(
+        server_details.return_value.addresses
+    )
+
+
+@patch("rabbit_consumer.openstack_api.get_server_details")
+def test_get_server_networks_no_network(server_details, vm_data):
     """
     Tests that an empty list is returned when there are no internal networks
     """
```
meoflynn, line 143: Does this test also cover the case when there are no internal and no services networks?

khalford: Yes it does cover that, I've removed the public network placeholder from the dictionary so it is more clear

meoflynn (dismissed):

meoflynn:
Looks good to me, although one of the github actions check seems to have failed and needs to be fixed first

DavidFair (approved):
LGTM, and the tests are significantly better than they were before

meoflynn (approved):

PR #135, opened 2024-03-12
ENH: Initial commit

khalford:
Created a Python script to remove the human error from manipulating SSL certificates.  The script will reverse the chain order of an SSL certificate and if you want , it can add the private key to the top of the chain.


anish-mudaraddi (approved):
Great work! Code is nice and readable! :+1:

PR #140, opened 2024-05-28
Cloud chatops refactor

khalford:
This PR adds the Cloud ChatOps application which is unfinished.


RWhiting78 (dismissed):

cloud_chatops/src/get_github_prs.py
```diff
@@ -0,0 +1,109 @@
...
+        :return: A dictionary of repos and their PRs.
+        """
+        responses = {}
+        for repo in self.repos:
+            url = f"https://api.github.com/repos/{self.owner}/{repo}/pulls"
+            response = self.get_http_response(url)
+            responses[repo] = response
+        return responses
+
+    def get_http_response(self, url: str) -> List[Dict]:
```
anish-mudaraddi, line 48: This class is getting too big imo - you have a lot of public methods here that aren't directly related to getting GitHub repos and can be moved to a more generic "Handler" Class. 

We want to make the class have a single responsibility so its easier to debug

Here's how I would go about this:

`GetGitHubPRs` - should have one publicly facing function called `run` - it's responsibility should be associated with getting repos from GitHub

Create a class like `HTTPHandler` with this `get_http_response` as the publicly facing function - its responsibility is handling http responses.  This class should be instantiated and used by `GetGitHubPRs` class (known as composition in OOP)

cloud_chatops/src/get_github_prs.py
```diff
@@ -0,0 +1,109 @@
...
+        :return: Dictionary of responses.
+        """
+        culled_responses = self.remove_empty_response(responses)
+        for repo, response in culled_responses.items():
+            if isinstance(response, dict):
+                responses[repo] = [response]
+        return responses
+
+    @staticmethod
+    def remove_empty_response(
```
anish-mudaraddi, line 74: this method should be private

cloud_chatops/src/get_github_prs.py
```diff
@@ -0,0 +1,109 @@
...
+        for repo, response in responses.items():
+            if not response:
+                to_remove.append(repo)
+
+        for i in to_remove:
+            del responses[i]
+        return responses
+
+    @staticmethod
+    def validate_response(response: requests.get, url: str) -> None:
```
anish-mudaraddi, line 93: this method should be private

cloud_chatops/src/pr_reminder.py
```diff
@@ -0,0 +1,205 @@
...
+        """
+        This method iterates through each PR and calls the post method for them.
+        :param mention: To mention the users or not
+        :param prs: A list of PRs from GitHub
+        :param reminder_message: The reminder message object
+        """
+        prs_found = False
+        for repo in prs.values():
+            for pr in repo:
+                info = {
```
anish-mudaraddi: if we set `GetGitHubPRs.run()` to output a list of dataclasses with the information already parsed into the way we want we could potentially remove the need for this

cloud_chatops/src/pr_reminder.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    This class handles the Slack posting.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.repos = get_repos()
+        self.client = WebClient(token=get_token("SLACK_BOT_TOKEN"))
+        self.slack_ids = get_user_map()
+        self.prs = GetGitHubPRs(get_repos(), "stfc").run()
```
anish-mudaraddi, line 22: To make this easier - we could potentially have`GetGitHubPRs.run()` return a list of dataclasses - where each dataclass contains information about a PR

cloud_chatops/src/pr_reminder.py
```diff
@@ -0,0 +1,205 @@
...
+        self.channel = "temp-chatops"
+
+    def run(self, mention, channel=None) -> None:
```
anish-mudaraddi, line 25: this should be the only public facing method in the class - all others should be private since they should only be used internally

anish-mudaraddi: Rule of thumb is if you want to integrate this class into other automation code - what methods should make up the API and what should we not let other code outside the class access

cloud_chatops/src/pr_reminder.py
```diff
@@ -0,0 +1,205 @@
...
+        :param reminder: The thread message to send under.
+        """
+        self.client.chat_postMessage(
+            text="There are no outstanding pull requests open.",
+            channel=reminder.data["channel"],
+            thread_ts=reminder.data["ts"],
+            unfurl_links=False,
+        )
+
+    def check_pr(self, info: Dict) -> Dict:
```
anish-mudaraddi: this class is also getting too big - you have this doing 2 things, validating the PR and posting it - which should be two separate classes same as above

anish-mudaraddi: You should break the class up into 

`PRMessageBuilder` - which checks various collected PRs, validates and builds a message to send
`PRMessagePoster` - which takes a list of PR messages and posts them as a single slack message.

anish-mudaraddi (changes requested):
Here are some suggestions of how I would refactor your code to make it more extensible.

anish-mudaraddi (approved):

PR #142, opened 2024-06-12
Extending classes no dir

khalford:
This branch extends the current classes and improves the code readability using dataclasses.

The pylint errors here are fixed in #143

### DIFF
--- a/cloud_chatops/src/get_github_prs.py
+++ b/cloud_chatops/src/get_github_prs.py
@@ -11,6 +11,8 @@ from custom_exceptions import RepoNotFound, UnknownHTTPError, BadGitHubToken
 
 
 class GetGitHubPRs:
+    # pylint: disable=R0903
+    # Disabling this as in the future there is likely to be more public functions.
     """
     This class handles getting the open PRs from the GitHub Rest API.
     """
@@ -49,13 +51,15 @@ class GetGitHubPRs:
     def _responses_to_dataclasses(responses: List[Dict]) -> List[PrData]:
         responses_dataclasses = []
         for pr in responses:
-            responses_dataclasses.append(PrData(
-                        pr_title=f"{pr['title']} #{pr['number']}",
-                        user=pr["user"]["login"],
-                        url=pr["html_url"],
-                        created_at=pr["created_at"],
-                        draft=pr["draft"],
-                    ))
+            responses_dataclasses.append(
+                PrData(
+                    pr_title=f"{pr['title']} #{pr['number']}",
+                    user=pr["user"]["login"],
+                    url=pr["html_url"],
+                    created_at=pr["created_at"],
+                    draft=pr["draft"],
+                )
+            )
         return responses_dataclasses
 
 
@@ -63,16 +67,17 @@ class HTTPHandler:
     """
     This class makes the HTTP requests to the GitHub REST API.
     """
+
     def make_request(self, url: str) -> List[Dict]:
         """
         This method gets the HTTP response from the URL given and returns the response as a list.
         :param url: The URL to make the HTTP request to.
         :return: List of PRs.
         """
-        response = self._get_http_response(url)
+        response = self.get_http_response(url)
         return [response] if isinstance(response, dict) else response
 
-    def _get_http_response(self, url: str) -> List[Dict]:
+    def get_http_response(self, url: str) -> List[Dict]:
         """
         This method sends a HTTP request to the GitHub Rest API endpoint and returns all open PRs from that repository.
         :param url: The URL to make the request to
@@ -97,7 +102,9 @@ class HTTPHandler:
                     "Your GitHub api token is invalid. Check that it hasn't expired."
                 )
             case 404:
-                raise RepoNotFound(f'The repository at the url "{response.url}" could not be found.')
+                raise RepoNotFound(
+                    f'The repository at the url "{response.url}" could not be found.'
+                )
 
             case _:
                 raise UnknownHTTPError(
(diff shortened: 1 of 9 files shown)


meoflynn:
Does this PR require #143 to be merged first?

cloud_chatops/src/pr_reminder.py
```diff
@@ -1,175 +1,168 @@
...
     """
     This class handles the Slack posting.
     """
 
-    def __init__(self):
-        super().__init__()
-        self.repos = get_repos()
-        self.client = WebClient(token=get_token("SLACK_BOT_TOKEN"))
+    def __init__(self, mention=False):
+        self.channel = "C06U37Y02R4"  # STFC-cloud: dev-chatops
```
anish-mudaraddi, line 21: you should probably make this a CONSTANT_VARIABLE and define it at the top or better yet have it as a parameter in the constructor set to this as default. This allows us to change it more easily

cloud_chatops/src/pr_reminder.py
```diff
@@ -181,12 +174,41 @@ def get_real_name(self, username: str) -> str:
...
         :param user: GitHub username to check for
         :return: Slack ID or GitHub username
         """
-        if user in self.slack_ids:
-            return self.slack_ids[user]
+        if user not in self.slack_ids:
+            # If the PR author is not in the Slack ID mapping
+            # then we set the user to mention as David Fairbrother
+            # as the team lead to deal with this PR.
+            user = "U01JG0LKU3W"
```
anish-mudaraddi, line 187: another one where it would be best to have this as a parameter to the method - something like `default_user_id` with this as the default

cloud_chatops/src/pr_reminder.py
```diff
@@ -181,12 +174,41 @@ def get_real_name(self, username: str) -> str:
...
+        This method checks if the PR is older than 6 months.
+        :param time_created: The date the PR was created.
+        :return: PR older or not.
+        """
+        opened_date = datetime.fromisoformat(time_created).replace(tzinfo=None)
+        datetime_now = datetime.now().replace(tzinfo=None)
+        time_cutoff = datetime_now - timedelta(days=30 * 6)
+        if opened_date < time_cutoff:
+            return True
+        return False
```
anish-mudaraddi, line 204: return 
```suggestion
        return opened_date < time_cutoff
```

cloud_chatops/src/get_github_prs.py
```diff
@@ -53,57 +85,28 @@ def get_http_response(self, url: str) -> List[Dict]:
...
         :param response: The response to check.
-        :param url: The url that was not found.
         """
-        if response.status_code == 401:
-            raise BadGitHubToken(
-                "Your GitHub api token is invalid. Check that it hasn't expired."
-            )
-        if response.status_code == 404:
-            raise RepoNotFound(f'The repo at the url "{url}" could not be found.')
+        match response.status_code:
```
anish-mudaraddi, line 97: nice to see usage of case statements :) - though its quite a new thing in Python

Since you're containerizing this script you can leave as is. But if you want it to run on older versions of python - you can use a dictionary like this:

```python
    { 
        200: lambda: pass
        401: lambda: raise BadGitHubToken(...),
        404: lambda: raise RepoNotFound(...)
    }.get(response.status_code, lambda: UnknownHTTPError(...))()
```

But I do agree the case statement is much cleaner

anish-mudaraddi (changes requested):
small nitpicks

cloud_chatops/src/get_github_prs.py
```diff
@@ -22,29 +25,58 @@ def __init__(self, repos: List[str], owner: str):
...
+        responses = []
         for repo in self.repos:
             url = f"https://api.github.com/repos/{self.owner}/{repo}/pulls"
-            response = self.get_http_response(url)
-            responses[repo] = response
+            responses += self._http_handler.make_request(url)
         return responses
 
+    @staticmethod
+    def _responses_to_dataclasses(responses: List[Dict]) -> List[PrData]:
```
anish-mudaraddi, line 51: copied over from #143 - mising docstring and function could be named better - something like `parse_responses` or `build_responses`

cloud_chatops/src/get_github_prs.py
khalford: That's useful to know, I haven't seen it before.

anish-mudaraddi (approved):

cloud_chatops/src/pr_reminder.py
```diff
@@ -1,175 +1,165 @@
...
+                f' You can use "/invite @Cloud ChatOps" to invite the app to your channel.'
+            )
+
+
+class PRMessageBuilder:
+    """This class handles constructing the PR messages to be sent."""
+
+    # pylint: disable=R0903
+    # Disabling this as there only needs to be one entry point.
+    def __init__(self, mention, default_user_id="U01JG0LKU3W"):
```
meoflynn, line 130: Who is the default user?

meoflynn (changes requested):

cloud_chatops/src/pr_reminder.py
```diff
@@ -1,175 +1,166 @@
...
+    # Disabling this as there only needs to be one entry point.
     """
     This class handles the Slack posting.
     """
 
-    def __init__(self):
-        super().__init__()
-        self.repos = get_repos()
-        self.client = WebClient(token=get_token("SLACK_BOT_TOKEN"))
+    def __init__(self, mention=False, channel="C06U37Y02R4"):
```
meoflynn, line 20: Which channel is this?

meoflynn (changes requested):

cloud_chatops/src/pr_reminder.py
khalford: I will add a comment in.

anish-mudaraddi (approved):

meoflynn (approved):

PR #144, opened 2024-06-20
add terraform script to make a private network

max-amb:
Adds a terraform script to create (and destroy) a private network within openstack!


meoflynn (dismissed):

terraform_provisioning/README.md
khalford, line 1: Here you will need to specify that the terraform script needs to be run with `--var-file=vars.tfvars` as this will pass through the user defined variables

khalford (changes requested):
One comment about documentation

khalford (approved):

meoflynn (approved):

PR #150, opened 2024-06-21
ENH: Added load balancing application in terraform

max-amb:
Added a terraform script to create load balancing with multiple vm's


terraform_provisioning/load_balancing_application/load_balancing_application.tf
```diff
@@ -0,0 +1,166 @@
...
+resource "openstack_lb_member_v2" "web_server_member" {
+  pool_id       = openstack_lb_pool_v2.web_servers_pool.id
+  count = length(openstack_compute_instance_v2.vm)
+  address = openstack_compute_instance_v2.vm[count.index].access_ip_v4
+  protocol_port = 80
+}
+
+# Creating the keypairs for the vms
+resource "openstack_compute_keypair_v2" "vm-keypair" {
+  name = "vm-keypair"
```
DavidFair, line 117: Nit: This should take a var for the name, the var could default to `vm-keypair`

terraform_provisioning/load_balancing_application/load_balancing_application.tf
```diff
@@ -0,0 +1,166 @@
...
+  connection {
+    type = "ssh"
+    user = "ubuntu"
+    host = openstack_compute_instance_v2.bastion.access_ip_v4
+    port = 22
+  }
+
+  filename = "/home/ubuntu/.ssh/id_rsa"
+  content = openstack_compute_keypair_v2.vm-keypair.private_key
+}
```
DavidFair, line 131: Could we drop this? You might wipe somebodies key pair doing it

Instead we should look for the public key and upload that directly - if it doesn't exist then...to honest if they can't generate an SSH key I'm surprised they're doing ansible :sweat_smile:

terraform_provisioning/load_balancing_application/load_balancing_application.tf
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    provisioner "file" {
+        source = "id_rsa"
+        destination = "/home/ubuntu/.ssh/id_rsa"
+    }
+
+    depends_on = [ openstack_lb_member_v2.ssh_member ]
+}
+*/
```
DavidFair, line 150: We can drop this then

terraform_provisioning/load_balancing_application/load_balancing_application.tf
```diff
@@ -0,0 +1,166 @@
...
+  flavor_name = var.flavor_name
+  key_pair = openstack_compute_keypair_v2.vm-keypair.name
+  name = format("vm-%d", count.index)
+
+  network {
+    uuid = openstack_networking_network_v2.private_network.id
+  }
+
+  # Due to weird issue faced with terraform creating vm's before subnets are made
+  depends_on = [ openstack_networking_subnet_v2.subnet ]
```
DavidFair, line 165: We should put a depends on in the LB member for these VMs too, else it will create LB listeners before the VMs come up

terraform_provisioning/load_balancing_application/vars.tfvars
```diff
@@ -0,0 +1,6 @@
+external_network_id = ""
+floating_ip = ""
+image_name = ""
+flavor_name = ""
```
DavidFair, line 4: I'd set this by default to a Ubuntu nogui image and l3.nano

terraform_provisioning/load_balancing_application/vars.tfvars
```diff
@@ -0,0 +1,6 @@
+external_network_id = ""
```
DavidFair, line 1: This should be `External` I think it's applicable in all cases

DavidFair (commented):
Nice work, I'm really excited to get this in since it's something we've long needed!

Just some nit picks around the edges mostly to make it safer and easier if people are trying it out

terraform_provisioning/load_balancing_application/load_balancing_application.tf
```diff
@@ -0,0 +1,166 @@
...
+resource "openstack_networking_network_v2" "private_network" {
+  name           = "private_network"
+  admin_state_up = "true"
+}
+
+# Creates a subnet within our private network
+resource "openstack_networking_subnet_v2" "subnet" {
+  name       = "subnet"
+  network_id = openstack_networking_network_v2.private_network.id
+  cidr       = "10.0.0.0/24"
```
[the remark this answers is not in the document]
DavidFair: To be clear the line then turns into `cidr = "192.168.3.0/24" instead

terraform_provisioning/load_balancing_application/load_balancing_application.tf
```diff
@@ -0,0 +1,166 @@
...
+
+# Create a ssh listener for the bastion vm
+resource "openstack_lb_listener_v2" "ssh_listener" {
+  name = "bastion-ssh"
+  protocol        = "TCP"
+  protocol_port   = 22
+  loadbalancer_id = openstack_lb_loadbalancer_v2.loadbalancer.id
+  timeout_client_data = 600000
+  timeout_member_connect = 600000
+  timeout_member_data = 600000
```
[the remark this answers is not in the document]
max-amb: This is specially for the ssh listener to be able to hold prolonged connections but I will remove it from the `web_server_listener` listener to deal with the issue.

terraform_provisioning/load_balancing_application/load_balancing_application.tf
max-amb: We need the key to access the application vm's from the bastion vm, should I rename to something non-standard like `application_access_key`.

DavidFair: But we could set the user's key on the VMs then use ssh key forwarding?

meoflynn (approved):

khalford (changes requested):
Unfortunately, we did not have the time to go through this work in detail and so I have left comments about some best practice things that you should change.

khalford (changes requested):
Good work, much more useable/readable now. Only one comment.

khalford (approved):
LGTM

DavidFair (approved):

DavidFair:
Thanks for your great work Max!

PR #151, opened 2024-07-02
ENH: Add support for multiple rabbit hosts in con string

DavidFair:
We should be connecting to the rabbit MQ hosts directly, this is currently preventing us from consuming messages as we previously went through a LB which is now unavailable

- Support constructing a connection string which contains multiple host endpoints. 
  This avoids us going through an LB and matches the approach used by OpenStack more closely.

- TBC Update host endpoint to point directly to our HV's in the chart
- Update chart version to ensure the new app version is picked up


OpenStack-Rabbit-Consumer/rabbit_consumer/message_consumer.py
```diff
@@ -254,6 +254,31 @@
...
+
+    for host in config.rabbit_hosts.split(","):
+        host = host.strip()
+        connect_str += f"{config.rabbit_username}:{config.rabbit_password}@{host}:{config.rabbit_port},"
+
+    # Trim the trailing comma
+    connect_str = connect_str[:-1]
+
+    sanitised_connect_str = connect_str.replace(config.rabbit_password, "<password>")
+    logger.debug("Connecting to rabbit with: %s", sanitised_connect_str)
```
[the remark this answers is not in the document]
DavidFair: Dismissed as a false positive: we replace the password with a placeholder in the line before hence the `sanitised_connect_str`
We also have a dedicated test to ensure this placeholder replacement is performed correctly, and it has been manually checked too.

meoflynn (dismissed):

OpenStack-Rabbit-Consumer/rabbit_consumer/message_consumer.py
```diff
@@ -254,6 +254,31 @@ def on_message(message: rabbitpy.Message) -> None:
...
```
anish-mudaraddi, line 277: hmm - this just feels wrong to me. You're unnecessarily risking the password being leaked  

I'd prefer you construct a variable separate that we can then output
```suggestion
    connect_str, debug_str = "amqp://"
    
    for host in config.rabbit_hosts.split(","):
        host = host.strip()
        connect_str += f"{config.rabbit_username}:{config.rabbit_password}@{host}:{config.rabbit_port},"
        debug_str += f"<username>:<password>@{host}:{config.rabbit_port},"

    # Trim the trailing comma
    connect_str = connect_str[:-1]
    debug_str = debug_str[:-1]

    logger.debug("Connecting to rabbit with: %s", debug_str)
```

anish-mudaraddi (changes requested):

anish-mudaraddi (approved):

OpenStack-Rabbit-Consumer/tests/test_message_consumer.py
```diff
@@ -101,34 +102,82 @@ def test_on_message_accepts_event_types(message_event_type, consume, event_type)
...
+    Test that the password does not get logged
+    """
+    returned_str = generate_login_str(mocked_config)
+    logging.debug.assert_called_once()
+    logging_arg = logging.debug.call_args[0][1]
+    assert mocked_config.rabbit_password in returned_str
+
+    # Check that the password is not in the log message
+    assert mocked_config.rabbit_username in logging_arg
+    assert mocked_config.rabbit_password not in logging_arg
```
meoflynn, line 164: As we have updated the method in the latest commit, do we still need this unit test?

meoflynn (changes requested):

OpenStack-Rabbit-Consumer/tests/test_message_consumer.py
meoflynn: Discussed in Slack channel - this unit test to be kept as an additional safety net

meoflynn (approved):

anish-mudaraddi (dismissed):

anish-mudaraddi (approved):

meoflynn (approved):

PR #154, opened 2024-08-13
BUG: fix issue where make fails with error status 400

anish-mudaraddi:
This failure should be ignored since the VM gets created fine in aquilon.

We should work out why this is the case later


anish-mudaraddi:
not going to fix that code cov issue - since its a test to see if an error is not raised - we don't need to test that the test fails :)

khalford (approved):
LGTM

meoflynn (approved):

PR #172, opened 2024-09-19
Cloud chatops docker secrets

khalford:
Created a docker deployment for Cloud Chatops. Uses GitHub Actions as CD to build and push dev/prod images. Uses watchtower to keep the Cloud Chatops container up to date.

Unit tests and code "de-duplication" to be done/fixed in future PRs


RWhiting78 (approved):

RWhiting78 (dismissed):

.github/workflows/cloud_chatops.yaml
```diff
@@ -1,16 +1,19 @@
...
     branches:
       - master
   pull_request:
     paths:
-      - "cloud_chatops/**"
       - ".github/workflows/cloud_chatops.yaml"
+      - "cloud_chatops/src/**"
+      - "cloud_chatops/tests/**"
+      - "cloud_chatops/Dockerfile"
+      - "cloud_chatops/compose.yaml"
```
DavidFair, line 13: Why do we get so specific? 
IMO: It should be anything in this folder triggers the workflow (better to be safe rather than sorry)

.github/workflows/cloud_chatops.yaml
```diff
@@ -25,17 +28,80 @@ jobs:
...
+          password: ${{ secrets.STAGING_HARBOR_TOKEN }}
+
+      - name: Build and push to staging project
+        uses: docker/build-push-action@v6
+        with:
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          push: true
+          context: "{{defaultContext}}:cloud_chatops"
+          tags: "harbor.stfc.ac.uk/stfc-cloud-staging/cloud-chatops"
```
DavidFair, line 80: This needs some sort of tag with a version, else it will push to `latest`
We use the SHA elsewhere

.github/workflows/cloud_chatops.yaml
```diff
@@ -25,17 +28,80 @@ jobs:
...
+
+      - name: Build and push on version change
+        uses: docker/build-push-action@v6
+        if: steps.release_updated.outputs.version == 'true'
+        with:
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          push: true
+          context: "{{defaultContext}}:cloud_chatops"
+          tags: "harbor.stfc.ac.uk/stfc-cloud/cloud-chatops"
```
DavidFair, line 107: Ditto about having a version here too, this should pull from a `version.txt`
That way your `if:` above can simply do something like `if: file_changed(cloud_chatops/version.txt)` (in pseudo-code)

cloud_chatops/.pylintrc
```diff
@@ -4,5 +4,7 @@
 max-line-length=120
 
 # Disable various warnings:
+# E0401 Ignored as imports work and will be tested when deployed in dev
+# R0801 Ignored to be fixed in future PRs
```
DavidFair, line 8: Nit: This should say:
`# R0801 - Duplicate Code - Ignored to be fixed in future PRs`

cloud_chatops/.pylintrc
```diff
@@ -4,5 +4,7 @@
 max-line-length=120
 
 # Disable various warnings:
+# E0401 Ignored as imports work and will be tested when deployed in dev
```
DavidFair, line 7: This needs to be fixed, it indicates a problem with the code structure
It's especially useful when we upgrade Python major versions and libraries end up in the Python std lib

cloud_chatops/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3
+WORKDIR /opt/cloud_chatops
```
DavidFair, line 2: Under the docs:
https://hub.docker.com/_/python

They point people to `/usr/src/app` so this looks suspect? Any reason we've broken convention in this image

cloud_chatops/compose.yaml
DavidFair, line 1: Should be `docker-compose.yml` as per convention

cloud_chatops/compose.yaml
```diff
@@ -0,0 +1,11 @@
+services:
+  cloud_chatops:
+    image: harbor.stfc.ac.uk/stfc-cloud/cloud-chatops
+    volumes:
+      - ./cloud_chatops_secrets/:/opt/cloud_chatops_secrets/
```
DavidFair, line 5: This should either use an absolute path or a dedicated Docker Compose volume

Because it's relative the effects can change (e.g. if you run `docker-compose up` from another folder)

cloud_chatops/pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-pythonpath = lib
+pythonpath = src
```
DavidFair, line 2: Shouldn't this be simply `.` or blank? Since we're now missing the lib folder
I would have thought the `from src import X` in `src/__init__.py` does this for us?

DavidFair (changes requested):

.github/workflows/cloud_chatops.yaml
khalford: 👍

cloud_chatops/Dockerfile
khalford: No reason, I did not know (look) that it was convention.

cloud_chatops/pytest.ini
khalford: Not sure I understand what you mean here. pythonpath is used to add the dirs to the python search path for imports. Hence src makes sense here as its the root module of code.

DavidFair (approved):

DavidFair (approved):

DavidFair (approved):

RWhiting78 (approved):

PR #180, opened 2024-10-03
Added scheduled job to build rabbit consumers and increment patch

Dmitry-Popovichev:
Pulls the latest container dependencies on a weekly basis. Created a bash script to increment the patch each time it runs. Created a separate yaml file for the scheduling so that the version doesn't increment on a pull or push. Commits the version.txt file to master.

### DIFF
--- a/.github/workflows/bin/version_increment.sh
+++ b/.github/workflows/bin/version_increment.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+
+#reads the version.txt file and sets the version to the current version
+version=`cat ~0/OpenStack-Rabbit-Consumer/version.txt`
+
+# cuts the $version variable into major, minor and patch numbers, removing the fullstop
+major=$(echo $version | cut -f1 -d.)
+minor=$(echo $version | cut -f2 -d.)
+patch=$(echo $version | cut -f3 -d.)
+
+#increments the patch by 1
+patch=$((patch + 1))
+
+#concatenate the version
+newversion="$major.$minor.$patch"
+
+#overwrites the version.txt file with new new version
+printf "$newversion" > ~0/OpenStack-Rabbit-Consumer/version.txt
--- a/.github/workflows/rabbit_consumer_schedule.yaml
+++ b/.github/workflows/rabbit_consumer_schedule.yaml
@@ -1,0 +1,120 @@
+name: Rabbit Consumer Schedule
+
+# Incrementing the patch version on a schedule to pull latest containers dependencies.
+
+
+on:
+  schedule: 
+      - cron: "12 10 * * TUE" # Every Tuesday at 10.12am
+
+jobs:
+  test_and_lint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10"]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "pip"
+      - name: Install dependencies
+        run: |
+          sudo apt-get update --fix-missing
+          python -m pip install --upgrade pip
+          # Required for requests-kerberos
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -yq install libkrb5-dev
+          pip install -r OpenStack-Rabbit-Consumer/requirements.txt
+          pip install -r OpenStack-Rabbit-Consumer/requirements-test.txt
+      
+      - name: Run tests
+        # Using Python3 to launch the module sets up the Python path for us
+        run: cd OpenStack-Rabbit-Consumer && python3 -m coverage run -m pytest .
+      
+      - name: Analyse with pylint
+        run: |
+          cd OpenStack-Rabbit-Consumer && pylint $(git ls-files '*.py')
+
+      - name: Prepare coverage
+        run: |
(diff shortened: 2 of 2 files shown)


RWhiting78 (approved):

.github/workflows/rabbit_consumer_schedule.yaml
```diff
@@ -0,0 +1,120 @@
+name: Rabbit Consumer Schedule
+
+# Incrementing the patch version on a schedule to pull latest containers dependencies.
+
+
+on:
+  schedule: 
+      - cron: "12 10 * * TUE" # Every Tuesday at 10.12am
```
anish-mudaraddi, line 8: why 10.12am? :)

.github/workflows/rabbit_consumer_schedule.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+      - name: Login to Harbor
+        uses: docker/login-action@v2
+        with:
+          registry: harbor.stfc.ac.uk
+          username: ${{ secrets.HARBOR_USERNAME }}
+          password: ${{ secrets.HARBOR_TOKEN }}
+
+        #increments ONLY the patch version using the version_incrememnt.sh script and commits the version.txt file to master
+      - name: Increment Version
```
anish-mudaraddi, line 93: Is there a way we only increment version when our dependency versions change? 
Set here. https://github.com/stfc/SCD-OpenStack-Utils/blob/master/OpenStack-Rabbit-Consumer/requirements.txt 

To do this - we need to put versions against each of our dependencies which dependabot can update for us

.github/workflows/rabbit_consumer_schedule.yaml
anish-mudaraddi: Also I'm not sure we need a schedule? 
Why not make it so this workflow runs when a PR is merged into master instead? 
Like a dependabot PR

.github/workflows/rabbit_consumer_schedule.yaml
anish-mudaraddi: This makes sure that we're not updating the patch version with no changes being made

.github/workflows/bin/version_increment.sh
jacob-ward, line 1: Advised to use shellcheck in the future - only flagged issues are advisories this time so will approve anyway

.github/workflows/rabbit_consumer_schedule.yaml
```diff
@@ -0,0 +1,120 @@
...
+        uses: docker/build-push-action@v3
+        with:
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          push: true
+          context: "{{defaultContext}}:OpenStack-Rabbit-Consumer"
+          tags: "harbor.stfc.ac.uk/stfc-cloud-staging/openstack-rabbit-consumer:${{ steps.commit_sha.outputs.sha_short }}"
+
+      - name: Inform of tagged name
+        run: echo "Image published to harbor.stfc.ac.uk/stfc-cloud-staging/openstack-rabbit-consumer:${{ steps.commit_sha.outputs.sha_short }}"
```
DavidFair, line 74: As this runs against `main` we should drop pushing the dev image, as it shouldn't fire

DavidFair: (Can be done as a follow-up tidying PR)

DavidFair:
Nice work overall, thanks :+1:

.github/workflows/rabbit_consumer_schedule.yaml
Dmitry-Popovichev: Roger that, I can clean that up today :+1:

.github/workflows/rabbit_consumer_schedule.yaml
anish-mudaraddi: If changing it to only run when PR is merged - It should only increment the version if the file hasn't already been changed manually - something like this might happen if we make a major change and need to update the minor/major version

anish-mudaraddi (changes requested):

.github/workflows/rabbit_consumer_schedule.yaml
Dmitry-Popovichev: 10.12am was sort of random but I chose Tuesday in the morning so that if there are any problems we can catch them and they are not left for after the weekend or at the end of the week.

I think we chose to schedule it just to keep on top of the dependency updates without having to check ourselves. 

We also do still have the original rabbit_consumer.yaml file that runs the workflow when we merge a PR into master, this is a duplicate file solely for scheduling. If we do do a minor or major change, we'll have to update version accordingly ourselves. Hopefully that makes sense :)

.github/workflows/rabbit_consumer_schedule.yaml
Dmitry-Popovichev: Quite possibly, I didn't look into this but me and David did discuss ways to include conditionals depending on what files are being updated but decided against it as it could over complicate it for now. Something that we could definitely consider though.

DavidFair: To follow-up I chatted to Anish about this, unfortunately the main target for this work is https://github.com/stfc/SCD-OpenStack-Utils/blob/1991246795d074d043eb76e58b78a9eee454ff09/OpenStack-Rabbit-Consumer/Dockerfile#L1 where we want to grab any security patches (or OS patches) for python 3.10

Because a label in Docker isn't static (unlike a tag in source control) 3.10 could refer to a base image with September patches one month, then October patches the next week but still be labelled `3.10`

PR #195, opened 2025-02-27
Remove old docker compose set up and replace with Ansible

khalford:
Ansible is more robust and allows greater / easier configuration. This should also replace our Aquilon configuration for the Grafana servers


Dmitry-Popovichev (approved):
Thanks for the PR review meeting, everything I had got answered in there. The only thing is the adding some documentation for the staging file and set up  (which would be separate anyway) and maybe hard coding the encrypted secrets.

DavidFair (approved):

PR #201, opened 2025-03-20
Remove domain variable

khalford:
Removing the domain variable allows us to run the playbook against multiple hosts. The domain variable is replaced with the inventory hostname. This means we have to put domains in the inventory file and not ip addresses


dev-0pz (approved):
LGTM

PR #207, opened 2025-03-26
Add terraform code to ChatOps

khalford:
Adding infrastrcuture setup in terraform. Creation of networking and compute resources. Also associated the floating IP to the load balancer as a bastion host


Chris-green-stfc (dismissed):

chatops_deployment/terraform/modules/compute/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  key_pair        = openstack_compute_keypair_v2.bastion_keypair.name
+  security_groups = ["default", var.loadbalancer_secgroup.name]
+
+  network {
+    name = var.private_network.name
+  }
+}
+
+data "openstack_networking_port_v2" "loadbalancer_port" {
+  fixed_ip = openstack_compute_instance_v2.loadbalancer.network[0].fixed_ip_v4
```
Neonbluestoplight, line 71: Why the [0]? 

Unless I am misunderstanding `openstack_compute_instance_v2.loadbalancer.network` seems to point to line 65 and that is just a list with name so you are adding fixed_ip and network_id to the name variable of the network?

chatops_deployment/terraform/variables.tf
```diff
@@ -0,0 +1,14 @@
+variable "deployment" {
+  type        = string
```
Neonbluestoplight, line 2: nitpick so not necessary to change: whitespace

Neonbluestoplight (approved):
Just one point I require clarification on but everything looks good enough

Neonbluestoplight (commented):
accidental

chatops_deployment/terraform/modules/compute/main.tf
khalford: > Curious: Why the [0]?

According to the Terraform documentation [here](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_instance_v2#network-2) virtual machines can be attatched to more than one network. As I am only attaching this VM to one network, the private one, I can use the array index 0 to get the first network in the list.

> data "openstack_networking_port_v2" "loadbalancer_port" {

This Terrraform data resource here retrieves the network port object that the VM was built on. We need this because in Terraform you have to associate the FIP to a network port specifically. Whereas in the Horizon GUI you just associate the FIP to an IPv4 address.

So to retrieve the network port we provide the `fixed_ip`, which is the IPv4 of the VM e.g. `192.168.100.2` and the `network_id`, which is the private network I created. We need to supply the `network_id` because multiple private networks can have the same subnets and therefore the same IPv4 addresses. One example of this is that every Kubernetes cluster builds nodes on the `192.168.x.y` subnet but on different networks.

chatops_deployment/terraform/variables.tf
khalford: Terraform has weird formatting. It likes all the variables equals to be on the same line like the below:
```
module "compute" {
  source                = "./modules/compute"
  grafana_secgroup      = module.networking.grafana_secgroup
  chatops_secgroup      = module.networking.chatops_secgroup
  prometheus_secgroup   = module.networking.prometheus_secgroup
  loadbalancer_secgroup = module.networking.loadbalancer_secgroup
  deployment            = var.deployment
  private_network       = module.networking.private_network
  floating_ip           = var.floating_ip
}
```
I don't know why, but I have updated the commit and reformatted all the files to use the terraform format

Neonbluestoplight: I see! That is strange but helpful to know

Neonbluestoplight (approved):

chatops_deployment/terraform/modules/compute/main.tf
```diff
@@ -0,0 +1,79 @@
...
+provider "openstack" {
+  cloud = "openstack"
+}
+
+resource "openstack_compute_keypair_v2" "bastion_keypair" {
+  name       = "bastion-keypair-${var.deployment}"
+  public_key = file("bastion-key.pub")
+}
+
+resource "openstack_compute_instance_v2" "grafana" {
```
anish-mudaraddi, line 20: why do you need three different compute instances for grafana/prometheus/chatops - why can't you run them all on one instance

khalford: Resiliancy, putting all 3 services on one VM would mean if that VM went down for any reason then you have no alerting (prometheus), monitoring (grafana) or application (chatops).

anish-mudaraddi: sure but why not have all 3 running on 3 VMs for maximum resiliency?

anish-mudaraddi: its fine for now - something to think about moving forwards

PR #209, opened 2025-04-28
ChatOps Deployment

khalford:
Adding Ansible configuration for the HAProxy load balancer and ChatOps hosts.

The ansible loadbalancer role installs HAProxy and sets up certbot to automatically generate the SSL certificate for the load balancer.
The ansible chatops role installs the docker.io package and runs the ChatOps container with the config and secrets files mapped as volumes.
There is also a change to the terraform deployment as we need to allow traffic to the docker metrics port for prometheus in the future


khalford:
Changed status to draft as the linting workflow was not running correctly on this PR. That is fixed in #210. Enabling the workflow to work will likely raise more problems which will need to be fixed before merging

RWhiting78 (approved):

chatops_deployment/ansible/roles/chatops/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
+---
+- name: Install docker
+  become: true
+  ansible.builtin.apt:
+    name: docker.io
+    update_cache: true
```
gmatthews20, line 6: You could add `state: latest` here to ensure that the latest version of the package is always installed

chatops_deployment/ansible/roles/chatops/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    name: ubuntu
+    groups: docker
+    append: true
+
+- name: Reset ssh connection to allow user changes to affect ansible user
+  ansible.builtin.meta: reset_connection
+
+- name: Start ChatOps container
+  community.docker.docker_container:
+    image: harbor.stfc.ac.uk/stfc-cloud-staging/cloud-chatops:b5ce2a8
```
gmatthews20, line 46: Just checking this should be pointing at a specific tag rather than latest?

chatops_deployment/ansible/roles/loadbalancer/tasks/certbot.yml
```diff
@@ -0,0 +1,98 @@
+- name: Install prerequisite packages
+  become: true
+  ansible.builtin.apt:
+    pkg:
+      - python3-venv
+      - libaugeas0
+    update_cache: true
```
gmatthews20, line 7: You could add `state: latest` here to ensure that the latest version of the package is always installed

chatops_deployment/ansible/roles/loadbalancer/tasks/certbot.yml
```diff
@@ -0,0 +1,98 @@
...
+
+- name: Install certbot
+  become: true
+  ansible.builtin.pip:
+    name: certbot
+    virtualenv: /opt/certbot
+    virtualenv_command: python3 -m venv
```
gmatthews20, line 14: You could add `state: latest` here to ensure that the latest version of the package is always installed

chatops_deployment/ansible/roles/loadbalancer/tasks/certbot.yml
```diff
@@ -0,0 +1,98 @@
...
+    minute: "3"
+    hour: "0,12"
+    day: "*"
+    job: "systemctl restart haproxy.service"
+
+- name: Restart HAProxy
+  become: true
+  ansible.builtin.systemd_service:
+    state: restarted
+    name: haproxy.service
```
gmatthews20, line 98: Does HAProxy need restarting each time the tasks are run or only when a cert is generated? If only when a cert is generated you could use a handler to only restart when the cert changes

chatops_deployment/ansible/roles/loadbalancer/tasks/haproxy.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Install HAProxy
+  become: true
+  ansible.builtin.apt:
+    name: haproxy
+    update_cache: true
```
gmatthews20, line 6: You could add `state: latest` here to ensure that the latest version of the package is always installed

gmatthews20 (changes requested):

gmatthews20 (approved):

RWhiting78 (approved):

PR #211, opened 2025-05-07
Add terraform play to ChatOps deployment

khalford:
This PR introduces playbooks to use Ansible to deploy the Terraform infrastructure instead of the Terraform CLI. The benefit of using Ansible to drive Terraform means the person deploying only needs to use one tool, Ansible, rather than Shell, Ansible and Terraform. 

Introduces 3 new plays instead of the single `site.yml`:

`deploy.yml`:
- Checks the SSH keys exist and are valid. If not it regenerates them.
- Deploys the infrastructure using an Ansible Terraform module.
- Edits the local user's known hosts file to trust the new or changed machines to avoid turning off `ANSIBLE_HOST_KEY_CHECKING`.
- Creates the `hosts.ini` file for the Ansible configuration

`destroy.yml`:
- Destroys all OpenStack resources and locally generated files

`configure.yml`:
- Does the configuration of the infrastructure to the hosts in the generated `hosts.ini` file.


Chris-green-stfc (approved):

chatops_deployment/ansible/roles/terraform/tasks/destroy.yml
```diff
@@ -0,0 +1,48 @@
...
+        path: hosts.ini
+        state: absent
+
+    - name: Touch hosts.ini file
+      ansible.builtin.file:
+        path: hosts.ini
+        state: touch
+        owner: "{{ ansible_env.USER }}"
+        group: "{{ ansible_env.USER }}"
+        mode: "0774"
```
anish-mudaraddi, line 33: why do you create a new empty hosts.ini file

khalford: When running `deploy.yml` it creates the OpenStack resources with Terraform and writes to the `hosts.ini` the IPs of the VMs. It then connects to each of those machines and graps their public key fingerprint to write to the local known hosts file. To be able to execute the known host tasks against the inventory in the same play as the writing to the file we must provide it with a blank file to begin with. Hence, touching the file after each destroy to make sure it is there.

anish-mudaraddi (approved):

Chris-green-stfc (approved):

PR #215, opened 2025-05-16
More monitoring for ChatOps

khalford:
I would suggest reviewers review commit-by-commit rather than the entire file diff

ChatOps services major deployment adds the following:

- Systemd exporter play to configure systemd-exporter for all hosts
- Add CAdvisor for ChatOps containers monitoring rather than the internal docker metrics endpoint
- Add Elastic stack VM in Terraform
 -Add Elasticsearch play
- Add Kibana play
- Add Logstash play
- Add filebeat to all hosts for all services to send their logs into logstash


chatops_deployment/ansible/roles/cadvisor/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+- name: Start CAsvisor container
+  community.docker.docker_container:
+    image: gcr.io/cadvisor/cadvisor
+    name: cadvisor
+    pull: always
+    state: started
+    ports:
+      - "8080:8080"
+    volumes:
+      - /:/rootfs:ro
```
DavidFair, line 27: We shouldn't mount the entire root dir, instead only dirs we need specifically to reduce the exfiltration surface

chatops_deployment/ansible/roles/elastic/tasks/elasticsearch_passwords.yml
```diff
@@ -0,0 +1,103 @@
...
+      ansible.builtin.shell: |
+        expect << EOF
+          spawn /opt/elasticsearch/bin/elasticsearch-reset-password -u elastic -s -i
+          expect -ex "Please confirm that you would like to continue \[y/N\]"
+          send "y\r"
+          expect -ex "Enter password for \[elastic\]:"
+          send "{{ elastic_password }}\r"
+          expect -ex "Re-enter password for \[elastic\]:"
+          send "{{ elastic_password }}\r"
+          expect eof
```
DavidFair, line 35: This should be an Ansible secret, rather than set dynamically at run time

DavidFair (changes requested):

chatops_deployment/ansible/roles/cadvisor/tasks/main.yml
khalford: I will look into that

chatops_deployment/ansible/roles/elastic/tasks/elasticsearch_passwords.yml
khalford: They are set as secrets here https://github.com/stfc/SCD-OpenStack-Utils/blob/more_monitoring/chatops_deployment/ansible/group_vars/monitoring/vars.yml

chatops_deployment/ansible/roles/filebeat/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+      changed_when: _.rc == 0
+
+    - name: Change permissions on binary files
+      become: true
+      ansible.builtin.file:
+        path: /opt/filebeat
+        state: directory
+        owner: root
+        group: root
+        recurse: true
```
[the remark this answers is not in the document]
khalford: Perhaps you can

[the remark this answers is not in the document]
khalford: I have managed to simplify this into 2 tasks. See new changes

PR #216, opened 2025-05-20
ENH: Add systemd exporter to hosts

khalford:
This enables us to put alerting in place for services that are running as a service on systemd. Such as Grafana, Elasticsearch and Logstash...

Adds the port 9558 as a rule to each VMs security groups.

### DIFF
--- a/chatops_deployment/ansible/configure.yml
+++ b/chatops_deployment/ansible/configure.yml
@@ -15,6 +15,14 @@
       tags:
         - chatops
 
+- name: Set up systemd exporters
+  hosts: all
+  remote_user: ubuntu
+  roles:
+    - role: systemd_exporter
+      tags:
+        - systemd_exporter
+
 - name: Configure Grafana
   hosts: grafana
   remote_user: ubuntu
--- a/chatops_deployment/ansible/roles/prometheus/files/rules.yml
+++ b/chatops_deployment/ansible/roles/prometheus/files/rules.yml
@@ -9,3 +9,12 @@ groups:
         annotations:
           summary: "ChatOps container is detected down on {{ $labels.instance }}"
           description: "Container has been not running for more than 10 seconds."
+
+      - alert: SystemdServiceDown
+        expr: systemd_unit_state{name=~"grafana-server.service|haproxy.service",state=~"failed|inactive"} == 1
+        for: 30s
+        labels:
+          severity: critical
+        annotations:
+          summary: "Systemd service {{ $labels.name }} on host {{ $labels.instance }} is in state {{ $labels.state }}."
+          description: "Systemd service has been in failed state for 30s."
--- a/chatops_deployment/ansible/roles/systemd_exporter/files/systemd-exporter.service
+++ b/chatops_deployment/ansible/roles/systemd_exporter/files/systemd-exporter.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Systemd Exporter Service
+Documentation=https://github.com/prometheus-community/systemd_exporter
+After=network-online.target
+
+[Service]
+User=systemd-exporter
+Group=systemd-exporter
+Restart=on-failure
+ExecStart=/opt/systemd-exporter/systemd_exporter \
+--web.telemetry-path="/"
+StandardOutput=append:/opt/systemd-exporter/systemd-exporter.log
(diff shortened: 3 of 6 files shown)


azahmd (approved):

chatops_deployment/ansible/configure.yml
```diff
@@ -15,6 +15,14 @@
       tags:
         - chatops
 
+- name: Set up systemd exporters
+  hosts: all
+  remote_user: ubuntu
```
DavidFair, line 20: I think you've used `remote_user` throughout, might be worth nipping this whilst we're here and putting

```
cat group_vars/all/vars.yml

ansible_user: ubuntu
```

To just set it for good, then override as required

chatops_deployment/ansible/roles/prometheus/files/rules.yml
```diff
@@ -9,3 +9,12 @@ groups:
         annotations:
           summary: "ChatOps container is detected down on {{ $labels.instance }}"
           description: "Container has been not running for more than 10 seconds."
+
+      - alert: SystemdServiceDown
+        expr: systemd_unit_state{name=~"grafana-server.service|haproxy.service",state=~"failed|inactive"} == 1
```
DavidFair, line 14: Should be `>= 1` as AFAIK this will stop calling out if two services fail?

chatops_deployment/ansible/roles/systemd_exporter/files/systemd-exporter.service
```diff
@@ -0,0 +1,16 @@
...
+Description=Systemd Exporter Service
+Documentation=https://github.com/prometheus-community/systemd_exporter
+After=network-online.target
+
+[Service]
+User=systemd-exporter
+Group=systemd-exporter
+Restart=on-failure
+ExecStart=/opt/systemd-exporter/systemd_exporter \
+--web.telemetry-path="/"
```
DavidFair, line 11: Nit: I thought this was a systemd argument called `web.telemetry` initially, any reason we don't simply

```
ExecStart=/opt/systemd-exporter/systemd_exporter --web.telemetry-path="/"
```

chatops_deployment/ansible/roles/systemd_exporter/files/systemd-exporter.service
```diff
@@ -0,0 +1,16 @@
...
+StandardOutput=append:/opt/systemd-exporter/systemd-exporter.log
+StandardError=append:/opt/systemd-exporter/systemd-exporter.log
```
DavidFair, line 13: Any reason we're not using normal logging locations e.g. `/var/log/systemd-exporter` as this would be way down on the list of places to look at?

chatops_deployment/ansible/roles/systemd_exporter/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+    mode: preserve
+    owner: systemd-exporter
+    group: systemd-exporter
+    remote_src: true
+
+- name: Copy systemd-exporter service file
+  become: true
+  ansible.builtin.copy:
+    src: systemd-exporter.service
+    dest: /etc/systemd/system/systemd-exporter.service
```
DavidFair, line 52: Another nit: it is a gray area but convention is anything not from package managers goes into `/usr/local`

In this case we could use `/usr/local/lib/systemd/system`, as per https://unix.stackexchange.com/a/367237 which is more correct since it's a package we've already put into `/opt` too

DavidFair (changes requested):

chatops_deployment/ansible/roles/prometheus/files/rules.yml
khalford: The metric for services being in any state "failed, inactive..." is binary.
0 Not in state
1 In state
So when state failed == 1, then we know that service is in that state

khalford: I need to update this line with other services anyways

chatops_deployment/ansible/roles/systemd_exporter/files/systemd-exporter.service
khalford: Yes we can do

DavidFair (approved):

Dmitry-Popovichev (approved):

PR #222, opened 2025-07-02
MAINT: Remove old repository not in use anymore

khalford:
Removing this repo as we don't use it anymore and it keeps our code cleaner.

This is tested to work in development.


khalford:
Just a demonstrative PR

PR #276, opened 2026-04-22
ENH: Allow user access to our Grafana instance

khalford:
This commit changes the deployment of our 2 Grafana instances. Going forward the production Grafana will be accessible by users. To make this the case, IRIS IAM users a part of the `stfc-cloud/users` will be able to authenticate. All datasources except those containing non-sensitive information have been removed. All dashboards except the slots available have been removed as they will no longer function.

The development instance will continue to contain all datasources and dashboards for the cloud team usage.


RyanH-STFC (approved):
Image

PR #280, opened 2026-04-29
BUG: Move dashboard into folder

khalford:
The "general" or top level directory in Grafana dashboards is readable only by admins. This means users cannot see the dashboards. Creating this "Public" subfolder allows it to be visible to editors and viewers.

### DIFF
--- a/grafana_monitoring/roles/grafana/templates/provision_dashboards.sh.j2
+++ b/grafana_monitoring/roles/grafana/templates/provision_dashboards.sh.j2
@@ -18,6 +18,7 @@ fi
 
 {% if inventory_hostname.startswith("grafana") %}
 for DIR in $(ls -l | grep -v cloud_dashboard.yaml | grep -v Slots-Available | grep -v total | awk -F' ' '{ print $NF }'); do rm -r $DIR; done
-mv Slots-Available/openstack_slots_available.json .
+mkdir "Public"
+mv Slots-Available/openstack_slots_available.json Public
 rm -r Slots-Available
 {% endif %}


grafana_monitoring/roles/grafana/templates/provision_dashboards.sh.j2
```diff
@@ -18,6 +18,7 @@ fi
 
 {% if inventory_hostname.startswith("grafana") %}
 for DIR in $(ls -l | grep -v cloud_dashboard.yaml | grep -v Slots-Available | grep -v total | awk -F' ' '{ print $NF }'); do rm -r $DIR; done
-mv Slots-Available/openstack_slots_available.json .
+mkdir "Public"
+mv Slots-Available/openstack_slots_available.json Public
```
jose-caballero, line 22: These 2 lines suggest that the file openstack_slots_available.json will always be first in the wrong place and needs to be moved to Public/ every time. Is that assumption correct? Or is there a chance the file is in the right path and that line will fail? 

I would suggest sending the stderr or mkdir to /dev/null, to avoid noise if Public/ already exist.

I personally don't like capital letters in directory names

khalford: Yes your assumption is correct. See line 16 which resets the directory to the Git repo main branch every time it runs. That means the Public directory should also be removed each time and will never conflict

I understand you don't like capital letters in directory names but this directory is used in the UI of Grafana which looks better. It is not used on the command line

image

jose-caballero: gotcha

jose-caballero (approved):

PR #282, opened 2026-05-13
Change RemoteWritePrometheus datasource setting

Dronesat:
Change timeout value from number to string according to grafana docs 
Change cacheLevel and timeInterval
Work to fix the timeout 504 issue on VM Power & Carbon dashboard

### DIFF
--- a/grafana_monitoring/roles/grafana/templates/grafana.ini.j2
+++ b/grafana_monitoring/roles/grafana/templates/grafana.ini.j2
@@ -32,3 +32,6 @@ log_level = info
 
 [security]
 admin_password="{{ grafana_admin_password }}"
+
+[dataproxy]
+timeout = 1200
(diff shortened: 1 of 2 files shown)


anish-mudaraddi (dismissed):

grafana_monitoring/roles/grafana/templates/grafana.ini.j2
```diff
@@ -28,7 +28,10 @@ role_attribute_path=contains(groups[*], 'stfc-cloud/admins') && 'Admin' || conta
...
 
 [log]
-log_level = info
+log_level = debug
 
 [security]
 admin_password="{{ grafana_admin_password }}"
+
+[dataproxy]
+timeout = 1200
```
khalford, line 37: Is this supposed to be 1200 or 120?

Dronesat: I'm just experimenting with different settings, so I thought I'd try a really high number

PR #287, opened 2026-07-29
Add cloud-tracker application to be worked on

dev-0pz:
docker not working yet and prometheus not working yet just there to template

uses cron jobs as db backups

this consists of the python scripts in app to be able to interact with the db via flash application and using apache as a reverse proxy on rocky.

plan is to use postgres to give metrics to prometheus 

no unit tests as of yet


dev-0pz:
got everything to run via docker images and containers rather than running off a vm httpd and systemd service

